PR #33, opened 2023-09-12
Claim Page: Claim Details Tab

ivanELEC:
## Description of change
Add data into claim details tab
Change some column and text classes across app to fit gov uk standards

## Link to relevant ticket

[Caseworker Build Claim Details Tab](https://dsdmoj.atlassian.net/browse/CRM457-434)


config/locales/en/claims.yml
```diff
@@ -35,13 +35,6 @@ en:
   claim_details:
     show:
       page_title: "Review criminal legal aid applications"
-  adjustments:
```
imtiazAhmd, line 38: Does this needs to be in ?

imtiazAhmd (approved):

PR #66, opened 2023-10-11
[CRM457-473] Your claims: Build landing page

imtiazAhmd:
## Link to relevant ticket

["Your claims": Build landing page + interactions](https://dsdmoj.atlassian.net/browse/CRM457-473)

## Screenshot

Screenshot 2023-10-12 at 08 12 42

## Todo

- [x] ASSESS NEXT CLAIM: IF user selects this button THEN the oldest unassigned claim from the “all claims list” should be shown to the user, and their name should be automatically allocated to that claim + the claim should be shown as being “Open” (See ticket: )

- [x] Check all claims: IF a user selects this THEN they will land on the “All Claims” tab

### DIFF
--- a/app/controllers/your_claims_controller.rb
+++ b/app/controllers/your_claims_controller.rb
@@ -5,5 +5,6 @@ class YourClaimsController < ApplicationController
     claims = Claim.your_claims
     claims = claims.map { |claim| BaseViewModel.build(:your_claims, claim) }
     @pagy, @claims = pagy_array(claims)
+    @next_claim = Claim.unassigned_claims.order(created_at: :desc).first
   end
 end
--- a/app/controllers/your_claims_controller.rb
+++ b/app/controllers/your_claims_controller.rb
@@ -1,3 +1,9 @@
+require 'pagy/extras/array'
 class YourClaimsController < ApplicationController
-  def index; end
+  include Pagy::Backend
+  def index
+    claims = Claim.your_claims
+    claims = claims.map { |claim| BaseViewModel.build(:your_claims, claim) }
+    @pagy, @claims = pagy_array(claims)
+  end
 end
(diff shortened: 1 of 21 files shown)


app/controllers/your_claims_controller.rb
```diff
@@ -1,3 +1,10 @@
+require 'pagy/extras/array'
 class YourClaimsController < ApplicationController
-  def index; end
+  include Pagy::Backend
+  def index
+    claims = Claim.your_claims
+    claims = claims.map { |claim| BaseViewModel.build(:your_claims, claim) }
+    @pagy, @claims = pagy_array(claims)
+    @next_claim = Claim.unassigned_claims.order(created_at: :desc).first
```
dwhenry, line 8: we should do this as if two please open the screen at the same time - wait 5 minutes, then click the link, they would both get the same claim. Add a new endpoint to select a claim, it will determine the new claim to assign, assign it and then redirect, this is a safer approach

app/view_models/v1/your_claims.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    def table_fields
+      [
+        { laa_reference: laa_reference, claim_id: id },
+        firm_name,
+        main_defendant_name,
+        date_created,
+        case_worker_name,
+        get_risk(risk)
```
dwhenry, line 45: don;t need to pass risk here, it is already visible in the scope you are passing it into

dwhenry (commented):
need to review claim assignment

app/controllers/your_claims_controller.rb
imtiazAhmd: Created a separate ticket to discuss the best way to handle assigning next claim to current user. Please feel free to add any details to the ticket.

[https://dsdmoj.atlassian.net/browse/CRM457-577](https://dsdmoj.atlassian.net/browse/CRM457-577)

dwhenry (approved):

dwhenry (approved):

PR #90, opened 2023-10-25
Add custom seed task to pipeline

ivanELEC:
So that we don't have to automatically run this task each time
It is ok for this to be in the pipeline because it won't duplicate existing claims, it just replaces


ivanELEC:
Not needed - we should eventually be relying on the app store to hold the data instead of forcing the data to upload from seeding

PR #99, opened 2023-10-27
Use Dev Auth in Development

ivanELEC:
## Description of change
Turn on Dev Auth in development environment so that real Azure authentication isn't being used

## Link to relevant ticket

[Caseworker Build: Enable Dev Auth in Development](https://dsdmoj.atlassian.net/browse/CRM457-651)

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,7 +1,7 @@
 feature_flags:
   dev_auth:
     local: <%= ENV.fetch("DEV_AUTH_ENABLED", false) %>
-    development: <%= ENV.fetch("DEV_AUTH_ENABLED", false) %>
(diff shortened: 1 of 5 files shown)


config/settings.yml
```diff
@@ -1,7 +1,7 @@
 feature_flags:
   dev_auth:
     local: <%= ENV.fetch("DEV_AUTH_ENABLED", false) %>
-    development: <%= ENV.fetch("DEV_AUTH_ENABLED", false) %>
+    development: <%= ENV.fetch("DEV_AUTH_ENABLED", true) %>
```
dwhenry, line 4: so this makes it impossible to be set to false... as if you set it to false it will return a string which is truthy... :) 

we are better off setting the ENV var if possible. :)

pking-moj (approved):

PR #108, opened 2023-11-01
[CRM457-630] Disbursements Edit Screen

imtiazAhmd:
## Description of change
Added Edit screen for disbursements from the Review and adjust tab
## Link to relevant ticket

[https://dsdmoj.atlassian.net/browse/CRM457-630](https://dsdmoj.atlassian.net/browse/CRM457-630)

## Screenshots of changes 
Screenshot 2023-11-08 at 15 21 13

##Todo:

- [x] Fix forms Save issue
- [x] Fix translation for the radio
- [x] Update rSpecs
- [x] Remove commented out line 26 in `disbursement_form` 
- [x] Add rSpecs tests
- [x] Update error translation
- [x] Add system test
- [x] Update 'explanation' to previous_explanation


app/lib/cost_calculator.rb
```diff
@@ -22,11 +22,7 @@ def letter_and_call_cost(object, scope)
     end
 
     def disbursement_cost(object)
-      if object.disbursement_type.value == 'other'
-        object.total_cost_without_vat
-      else
-        object.miles * object.pricing
-      end
+      object.provider_requested_total_cost_without_vat + object.vat_amount
```
dwhenry, line 25: so I think this will return the incorrect value for the provider requested total cost when the disbursement is disallowed as it is using the `current` value of the `vat_amount` when it should be the `provider_requested_vat_amount`

dwhenry (changes requested):
Potential issue with total calculation 

Can we also add a system spec (just copy fro what already exists as much as possible) - happy to help on this if you want. 

> NOTE: the testing is ignoring the fact turboframes needs javascript by directly loading the data pages)

app/lib/cost_calculator.rb
imtiazAhmd: Provider don't request the vat_amount for Disbursement items. It gets applied based on the items. In JSON we get 0 as vat_amount if item doesn't qualify or the calculated vat_amount if it does. Happy to have a chat if this doesn't make any sense

imtiazAhmd: system spec added

dwhenry: so the issue is we are using the provider requests total cost and the current vat_amount

This is inconsistent as when the caseworker has disallowed the disbursement the vat amount for the provider requested amount is adjusted to 0.

spec/controllers/disbursements_controller_spec.rb
```diff
@@ -27,4 +27,101 @@
...
+      end
+    end
+  end
+
+  context 'update' do
+    let(:claim) { instance_double(Claim, id: claim_id) }
+    let(:claim_id) { SecureRandom.uuid }
+    let(:form) { instance_double(DisbursementsForm, save:) }
+    let(:disbursement_id) { '1c0f36fd-fd39-498a-823b-0a3837454563' }
+    let(:disbursement) { instance_double(V1::Disbursement, id: disbursement_id, attributes: item) }
```
dwhenry, line 83: `attributes: item` this can;t be wrong based on what the next lines does.

It feels like the variable (let) names have gotten confessing - I would recommend a review and work out meaningful names if possible to avoid this.

spec/support/disbursement_system_helpers.rb
```diff
@@ -0,0 +1,11 @@
+module DisbursementSystemHelpers
+  def verify_disbursement_item
+    within('.govuk-table__row', text: 'Apples') do
```
dwhenry, line 3: not sure having this hear is helping, any reason for not leaving this inline in the test?

dwhenry (commented):
still have concerns about the cost calculator

ivanELEC (dismissed):

spec/support/disbursement_system_helpers.rb
imtiazAhmd: Removed this helper for now, we will revisit as part of cleanup.

spec/controllers/disbursements_controller_spec.rb
imtiazAhmd: Updated

imtiazAhmd:
> still have concerns about the cost calculator
@dwhenry We can have a quick call to discuss your concern

dwhenry (approved):
Can't see the code around not prepopulating the yes no button... assume it in a separate PR...

imtiazAhmd:
> Can't see the code around not prepopulating the yes no button... assume it in a separate PR...

Yes it would be separate PR

PR #118, opened 2023-11-08
Crm457 433: Work Item Edit Functionality

ivanELEC:
## Description of change
- Add page for editing work items
-  Add period component to use for TIme Spent
- Add events for updating a Work Item
- Add form for editing work items with option to remove uplift, change time spent and add an explaination for the change
- Add data for Uplift Allowed and Caseworker Allowed in Review and Adjust: Work Items tab

## Link to relevant ticket

[Work Items - Edit screen](https://dsdmoj.atlassian.net/browse/CRM457-628)
[Work Items - Event Stream](https://dsdmoj.atlassian.net/browse/CRM457-629)

NOTE: messages for error handling needs to be complete still - this will be done in a separate pr


app/helpers/application_helper.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  include FormBuilderHelper
```
dwhenry, line 4: ~have a look at the `form_builder.rb` class in the provider app as prefer that model of adding this helper. :)~

This shouldn't be needed (I don;t think)

app/view_models/v1/work_item.rb
```diff
@@ -1,24 +1,60 @@
 module V1
-  class WorkItem < BaseViewModel
+  class WorkItem < BaseWithAdjustments
+    LINKED_TYPE = 'work_items'.freeze
+    ID_FIELDS = %w[id].freeze
+
+    attribute :id, :string
     attribute :work_type, :translated
     # TODO: import time_period code from provider app
     attribute :time_spent
```
dwhenry, line 9: I would consider converting this to a period so that it is easier to render in the display. :)

dwhenry (approved):
Looking pretty good to me. :)

PR #152, opened 2023-11-23
CRM457-599 - Downloadable Evidence

pking-moj:
## Description of change

- Added AWS gem and initialiser
- Updated view model for evidence
- Get presigned URL for each evidence

## Link to relevant ticket

[](https://dsdmoj.atlassian.net/browse/)

### DIFF
--- a/app/controllers/supporting_evidences_controller.rb
+++ b/app/controllers/supporting_evidences_controller.rb
@@ -4,7 +4,20 @@ class SupportingEvidencesController < ApplicationController
     claim_summary = BaseViewModel.build(:claim_summary, claim)
     supporting_evidence = BaseViewModel.build(:supporting_evidence, claim, 'supporting_evidences')
 
+    generate_urls supporting_evidence
+
     @pagy, @supporting_evidence = pagy_array(supporting_evidence)
     render locals: { claim:, claim_summary: }
+  end
+
+  private
+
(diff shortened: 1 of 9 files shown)


app/controllers/supporting_evidences_controller.rb
```diff
@@ -4,7 +4,20 @@ def show
...
     render locals: { claim:, claim_summary: }
   end
+
+  private
+
+  def generate_urls(supporting_evidence)
+    supporting_evidence.each do |item|
+      item.download_url = S3_BUCKET
+                          .object(item.file_path)
+                          .presigned_url(:get, expires_in: 900,
```
dwhenry, line 19: to confirm this is 1.5min or 1.5 hours - if the first it doesn;t seem long enough

would also prefer this to be an const. :)

dwhenry (approved):
**Note a blocker but:**

Feel like this is missing a test around setting the `download_url` on the supporting evidence object...

Event the system test only has the hostname that I can see.

app/controllers/supporting_evidences_controller.rb
pking-moj: This is 15 mins (900 seconds) - happy to set this as a const and add a comment

imtiazAhmd (dismissed):

dwhenry (approved):

PR #155, opened 2023-11-24
Add additional check for uplifts to prevent nil coersion to Integer

ivanELEC:
## Link to relevant ticket

[Incorrect Uplift Value Shown in Change Work Item Form When No Uplift](https://dsdmoj.atlassian.net/browse/CRM457-741)

[Extra context here](https://mojdt.slack.com/archives/C05212WRK5J/p1700823135194779)


ivanELEC:
No longer needed - modified claim side to not send nil

PR #168, opened 2023-11-30
CRM-756 - Add feedback mailer and initial messages

pking-moj:
## Description of change

- Add feedback mailer for claim updates
- Add message builders for each state
- Add relevant testing
- Update config around this implementation

## Link to relevant ticket

[CRM-756](https://dsdmoj.atlassian.net/browse/CRM-756)

## Notes for reviewer

- Some values are not easily accessible so have set this up using what is available currently but this may need to change
- Current do not know how to pull the comment cleanly so have added the ability to pass this in however i would like to refactor this
- Have tested locally against notify to ensure templates are rendered correctly successfully. Doing this in an environment will require a change of API key type


imtiazAhmd (dismissed):

app/mailers/feedback_messages/feedback_base.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def recipient
+      raise NotImplementedError, "#{self.class} has not implemented method '#{__method__}'"
+    end
+
+    protected
+
+    def main_defendant
+      @claim.data['defendants'].find { |defendant| defendant['main'] == true }
```
patrick-laa, line 25: You _could_ shorten this to:
```suggestion
      @claim.data['defendants'].find { _1['main'] }
```
(The `== true` is redundant unless `main` contains non-boolean values like strings that should be evaluated as `false`, and given that each item in `defendants` is a defendant by implication, the `_1` shorthand instead of a named variable makes things more concise (although depending on how used you are to it, you may find it harder to read).

app/services/notify_app_store.rb
```diff
@@ -3,10 +3,12 @@ class NotifyAppStore < ApplicationJob
 
   def self.process(claim:)
     if ENV.key?('REDIS_HOST')
-      set(wait: Rails.application.config.x.application.app_store_wait.seconds).perform_later(claim)
+      set(wait: Rails.application.config.x.application.app_store_wait.seconds)
+        .perform_later(claim)
     else
       begin
         new.notify(MessageBuilder.new(claim:))
+        ClaimFeedbackMailer.notify(claim).deliver_later!
```
patrick-laa, line 11: Is there a timing issue here? The feedback email will be delivered almost instantaneously, but the provider app only updates itself from the app store periodically, so will this lead to providers being told about an update to their claim, and then going to look at the provider app to see the specifics and there being nothing there?

spec/mailers/claim_feedback_mailer_spec.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# rubocop:disable RSpec/MultipleMemoizedHelpers
+RSpec.describe ClaimFeedbackMailer, type: :mailer do
```
patrick-laa, line 6: I would love to see this functionality covered by an integration spec as well as these unit specs, to 'prove' that all the moving parts join up as expected - something that says effectively "Given there is a claim, when a caseworker clicks the button to approve it, then an email is triggered to the provider". I know the 'what's our test strategy' conversation is still in flight, though, so I don't want to rock the boat while we don't have an agreed convention.

patrick-laa (commented):
Three comments from me, all potentially ignoreable.

app/services/notify_app_store.rb
pking-moj: As it stands your right there is a timing issue. However there is a possibility that is being discussed/understood around delaying sending the email for a period of time to allow caseworkers to be able to change/undo their actions and recall the email. This isnt implemented now however so that we can check it is sent and that content is correct

spec/mailers/claim_feedback_mailer_spec.rb
pking-moj: I will look at adding this in before proceeding with more reviews. Its good we try and cover this now rather than miss it

app/mailers/feedback_messages/feedback_base.rb
ivanELEC: only a personal opinion and I'm happy to be overruled but my preference would be to not use the _1 shorthand to make the code more explicit

ivanELEC (approved):

ivanELEC (approved):

PR #189, opened 2023-12-15
Update README

jsugarman:
## Description of change

Update README

- Database config example added as this is not an idiomatic rails
way of defining the database

- A yarn install should typically use a frozen lock file to keep
things consistent between local developer envs and hosted envs.

- rails assets:precompile was required to get things working as no
`/node_modules/` dir exists is stored in the repo.

- Tweaked the snippet for adding users based on what worked
on UAT environment


jogilder (approved):
LGTM

PR #225, opened 2024-01-15
CRM457-942: Make space for prior authority logic

patrick-laa:
## Description of change
- Renamed `claims` table to `submissions` 
- Enforced use of `application_type` field to differentiate CRM4 and CRM7
- Moved CRM7 into a `Nsm` namespace
- Added a `PriorAuthority` namespace with POC code that reuses the BaseViewModel and BaseAdjustmentForm to allow viewing and editing additional costs

## Link to relevant ticket

[CRM457-942](https://dsdmoj.atlassian.net/browse/CRM457-942)

[CRM457-942]: https://dsdmoj.atlassian.net/browse/CRM457-942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/view_models/base_view_model.rb
```diff
@@ -5,17 +5,17 @@ class BaseViewModel
...
   class Builder
-    attr_reader :klass, :claim, :rows, :return_array
+    attr_reader :klass, :crime_application, :rows, :return_array
 
-    def initialize(class_type, claim, *nesting)
-      # TODO: Add mechanism to determine namespace dynamically
-      @klass = "NonStandardMagistratesPayment::V#{claim.json_schema_version}::#{class_type.to_s.camelcase}".constantize
-      @claim = claim
+    def initialize(class_type, crime_application, *nesting)
+      namespace = crime_application.is_a?(Claim) ? 'NonStandardMagistratesPayment' : 'PriorAuthority'
```
dwhenry: would prefer to use the `application_type` in the check over the class - better yet move it into the `CrimeApplication` class so that it can return the appropriate namespace.

app/view_models/base_view_model.rb
```diff
@@ -36,10 +36,11 @@ def build
     private
 
     def params(attributes)
-      claim.attributes
-           .merge(claim.data)
-           .merge(attributes, 'claim' => claim)
-           .slice(*klass.attribute_names)
+      key = crime_application.is_a?(Claim) ? 'claim' : 'application'
```
dwhenry: I feel like this would get confusing and would prefer just a generic key like `submitted` or `model`

app/controllers/prior_authority/additional_costs_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def update
+      crime_application = PriorAuthorityApplication.find(params[:application_id])
+      all_costs = BaseViewModel.build(:additional_cost, crime_application, 'additional_costs')
+      index = all_costs.index do |model|
+        model.id == params[:id]
+      end
+
+      item = all_costs[index]
```
dwhenry, line 23: would be interested in understanding why you use `index` over `detect`

dwhenry (approved):
I reviewed by commit and it all seems to make sense

Comments may be addressed in later commits, but mostly just small style inputs. :)

spec/factories/assignments.rb
```diff
@@ -1,6 +1,9 @@
 FactoryBot.define do
   factory :assignment do
     user factory: %i[caseworker]
-    claim
+    transient do
+      claim { nil }
+    end
+    crime_application { claim || create(:claim) }
```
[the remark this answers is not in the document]
patrick-laa: You're right, that was messy! I was bending over backwards a bit to please rubocop, which doesn't like `a { create(:b) }` in factories, but then I realised I was really just trying to avoid updating old test code that assumed that assignments had a `claim` attribute. There was only 1 place that actually needed updating, so when I did that, I was left with the nice and straightforward:

```ruby
 submission factory: :claim
 ```
 
which avoids the need for a transient at all.

app/view_models/base_view_model.rb
patrick-laa: Oh that's a great idea. Done.

app/view_models/base_view_model.rb
patrick-laa: Good thinking. Done.

app/controllers/prior_authority/additional_costs_controller.rb
patrick-laa: It's only because we need to have `index` as a variable in the views so that, as per the current designs, we can use it in the page title (e.g. "Additional Cost 1"). Since we needed to get the index anyway, I thought we might as well get that, then use it to retrieve the item, rather than go through the list of items twice, once for the item and once for the index.

(This is probably throwaway code - I'm sure when we come to build the form in earnest there will be more scrutiny of how additional cost items are referred to, and the `index + 1` approach may go out the window)

dwhenry (approved):
Only looked at the last commit as already approved the rest

PR #243, opened 2024-01-25
CRM457-1003: Add mechanism to self-assign prior authority applications

patrick-laa:
## Description of change
Assign an appropriate application to a caseworker when they press the big button, using the [urgency rules](https://dsdmoj.atlassian.net/browse/CRM457-987)

## Link to relevant ticket

[CRM457-1003](https://dsdmoj.atlassian.net/browse/CRM457-1003)

## Notes for reviewer
* I've added a lock as we expect volume to be high and we want the edge case of 2 caseworkers hitting the same button at the same time and both assigning the same case to themselves not to bite us (as the first caseworker would find themselves trying to edit a submission they were no longer assigned to
* The `unassigned` scope was adding default ordering that got in the way of our custom ordering. Since ordering was already being explicitly added when it was being called anyway, I've removed the default
* The new logic makes assumptions about the structure of the JSON blob. Until we have a formal schema, I've added something that will throw errors if we try to import a blob that doesn't meet these assumptions, so we have a way of at least noticing if there's a problem.


spec/system/prior_authority/assign_applications_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    # As the UI has not yet been built, the most we can do is demonstrate that an assignment has been made
+    expect(application.reload.assignments.first.user).to eq caseworker
+    expect(application.events.first).to be_an Event::Assignment
+  end
+
+  it 'copes if there is no application' do
+    click_on 'Assess next application'
+    expect(page).to have_content 'There are no applications waiting to be allocated'
+  end
+end
```
jsugarman, line 26: **nit**. Could use a context for 'when [no] applications exist', with a before hook and let for the application creation/setup. makes it a little clear and spec more easily extendible. Or we cross that bridge when we com to it.

spec/services/prior_authority/choose_application_for_assignment_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+           data: build(:prior_authority_data, service_type: 'pathologist', court_type: 'other'))
+    expect(described_class.call(user)).to eq central
+  end
+
+  it 'prefers earlier submisions' do
+    create(:prior_authority_application, created_at: 1.day.ago)
+    prior = create(:prior_authority_application, created_at: 2.days.ago)
+    expect(described_class.call(user)).to eq prior
+  end
+end
```
jsugarman, line 32: **just nits**, but personally i prefer contexts and setup outside the examples themselves

jsugarman (approved):
LGTM, couple of spec nits

PR #244, opened 2024-01-25
Use feature_flags.yml instead of settings.yml

ivanELEC:
## Description of change
Change feature flags library to use feature_flags.yml instead of settings.yml

## Link to relevant ticket
[Formalise Use of Feature Flags](https://dsdmoj.atlassian.net/browse/CRM457-955)


patrick-laa (commented):
I think there's maybe a little bit more to do here. Currently the settings/feature_flags yaml doesn't have a value for the UAT environment. Instead it loads it in from an ENV var which loads it in from a 'secret' which is actually just a value in a helm_deploy values file.

I thought the upshot of the meeting was that we would keep feature flags simple and not using ENV vars. If so, we should remove every reference to `devAuthEnabled` and `DEV_AUTH_ENABLED` and simply set the value we want for UAT in the feature flags yaml file, rather than in a different yaml file in the same codebase.

patrick-laa (approved):
LGTM

PR #271, opened 2024-02-13
(SPIKE) CRM457-1099: No database

patrick-laa:
## Description of change
* The goal is to avoid having a local copy of application data, so removing the "submissions" table entirely
* This means dynamically loading lists of submissions and individual submissions when individual pages need them
* With this in place, the assignment logic stopped making much sense to have in the CW app, so I moved that to the app store
* Without syncing, at least some of the event logic has to exist in the app store, and the neatest approach I could find was to move it *all* there
* So this PR removes the submissions, assignments and events tables and turns all responsibility for that logic over to the app store.
* The heart of the logic is in app/services/app_store_service.rb

## Link to relevant ticket

[CRM457-1099](https://dsdmoj.atlassian.net/browse/CRM457-1099)

[CRM457-1099]: https://dsdmoj.atlassian.net/browse/CRM457-1099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


patrick-laa:
Closing because we've decided not to pursue this for MVP

PR #278, opened 2024-02-16
CRM457-1062: per-branch temporary dev environments in cloud platform

patrick-laa:
## Description of change
- CI now spins up a dev release with its own URL, e.g.. this one is available at https://crm457-1062-tempor-nscc-caseworker-dev.cloud-platform.service.justice.gov.uk
- To avoid contaminating other environments our schema changes and data, we create a postgres pod to use as our database
- Deleting the environment is done as a GitHub action so that it can trigger even when the PR is closed without merging (note I can't test this bit as the workflow can't run until it exists in `main` - once it's there we can trigger it manually and even choose which branch's version to trigger)

## Link to relevant ticket

[CRM457-1062](https://dsdmoj.atlassian.net/browse/CRM457-1062)

[CRM457-1062]: https://dsdmoj.atlassian.net/browse/CRM457-1062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -362,26 +354,22 @@ workflows:
...
           requires:
             - build-to-ecr
-      - deploy-dev-crm4:
-          context: laa-assess-non-standard-magistrates-fee-dev-crm4
-          requires:
-            - hold-dev-crm4
+          filters:
+            branches:
+              ignore:
+                - main
```
jsugarman, line 364: do we also want a "main-dev" hosted env, for each repo. These "main-dev" envs, for each repo, would be the fallbacks for the dynamic branch and for other "main-dev" branches by default
i.e. 
**my-provider-branch** --> main-dev-app-store <-- main-dev-caseworker
main-dev-provider --> main-dev app store <-- **my-branch-caseworker**

merging to main would then both delete dynamic branch and automatically deploy to the main-dev host for the repo.

helm_deploy/values-dev.yaml
```diff
@@ -66,3 +66,22 @@ secrets:
...
+      memory: 256Mi
+    requests:
+      cpu: 250m
+      memory: 256Mi
+  auth:
+    database: nsm_assess_dev
+    enablePostgresUser: true
+    postgresPassword: postgres
+  postgresqlUsername: postgres
+  primary:
```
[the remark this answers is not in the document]
patrick-laa: Yup, that got rid of 5 lines of warnings. Thanks!

jsugarman (approved):
looks like a good starting point, especially as it works 😄.  aside from the couple of comments we may want to extend to give branches their own redis instances, rather than a shared elasticache instance.

Sharing an elasticache instance between all branches could result in race conditions and odd behaviour when attempting to excercise the full journey. I have personally looked into this on Civil Apply and other apps and for a couple of reasons removed alternative  mechanisms such as branch specific queue. The Sidekiq author suggests separate redis datasbases but default limits on numbers of databases and ensuring correct config per branch makes this painful IMO. Simplest approach that I used on apply was to use separate redis pods, as for postgresql.


once in, worth remembering we can remove the RDS instance from the CP namespace.

.circleci/config.yml
patrick-laa: That's a really good idea!

patrick-laa:
Merging to see what breaks. Fully agree with the point about Redis pods.

PR #293, opened 2024-02-22
CMR457-1108: remove assumption checker

jsugarman:
## Description of change
remove assumption checker

[crm457-1108](https://dsdmoj.atlassian.net/browse/CRM457-1108)

Unless we are going to keep this upto date with
all attributes from the PA payload, and possibly
include expected inclusion list validations, then
we need to remove it.


app/lib/prior_authority/application_assumption_checker.rb
```diff
@@ -1,30 +0,0 @@
-module PriorAuthority
-  class ApplicationAssumptionChecker
-    include ActiveModel::Model
-    include ActiveModel::Attributes
-
-    COURT_TYPES = %w[central_criminal_court other].freeze
-    SERVICE_TYPES = %w[pathologist other].freeze
```
patrick-laa, line 7: So this is basically the only assumption that the caseworker app actually relies on that isn't true. I think as part of removing this assumption checker we should fix the assumption. Namely, now that we know that `pathologist` is actually `pathologist_report`, PriorAuthority::ChooseApplicationForAssignment needs a one-line update.

If we merge this PR without making that change my worry is we'll forget to ever make that change because it will no longer be apparent that there's a discrepancy.

jsugarman: Not sure i follow but I have [added this PR](https://github.com/ministryofjustice/laa-assess-non-standard-magistrate-fee/pull/296) to go in ahead of it anyway.

The actual error that was raised when i submitted during manual treating was office_code attributes does not exist. If we are to ever validate the service types I would expect us to go and extract the entire from Provider app (or other source of truth) and us in here. 🤔 A lookup API for this data would be good in the long term.

jsugarman: just looked at the ChooseApplicationForAssignment, and will amend too 👍

patrick-laa: Sorry, I'm not being clear. The reason AssumptionChecker _actually_ exists is to document that PriorAuthority::ChooseApplicationForAssignment assumes that some specific states of a PA application are represented by some specific string values. 

Now, the AssumptionChecker is now deeply broken because it doesn't know about `office_code` and a bunch of other stuff, and we should probably therefore delete it as per this PR.

BUT if we delete it we lose the documentation about PriorAuthority::ChooseApplicationForAssignment's assumptions. And this is a problem because one of those assumptions is _wrong_. So before we remove the documentation, we should fix the wrong assumption, that way in the absence of documentation we won't forget about it.

So the update we need to make while we're aware of it is to stop PriorAuthority::ChooseApplicationForAssignment thinking that `pathologist` is a valid service_type. It isn't. It's been replaced by `pathologist_report`. We should update `PriorAuthority::ChooseApplicationForAssignment `'s logic to match, because otherwise `PriorAuthority::ChooseApplicationForAssignment` is subtly broken.

patrick-laa: Ah, sorry, our comments crossed over one another. Ignore my longwinded comment, you've already understood what I mean!

jsugarman: @patrick-laa ready for review now

patrick-laa (approved):

PR #297, opened 2024-02-22
CRM457-1168: Add PA standard header

patrick-laa:
## Description of change
Update the existing proof-of-concept PA models with real attributes
Create a shared application header/subnav that matches the prototype, allowing individual tab content to be built out in later PRs.

## Link to relevant ticket

[CRM457-1168](https://dsdmoj.atlassian.net/browse/CRM457-1168)

## Notes for reviewer
This should unblock CRM457-1169 and CRM457-992

Screenshot 2024-02-22 at 16 00 45

[CRM457-1168]: https://dsdmoj.atlassian.net/browse/CRM457-1168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/locales/en/prior_authority.yml
```diff
@@ -48,16 +51,120 @@ en:
     assignments:
       create:
         no_unassigned_applications: There are no applications waiting to be allocated
+    service_types:
+      ae_consultant: A&E consultant
+      accident_reconstruction: Accident reconstruction
+      accountant: Accountant
+      anaesthetist: Anaesthetist
+      animal_behaviourist: Animal behaviourist
```
jsugarman, line 59: i suppose there is no simple way we can avoid baking in knowledge of the individual service types. long term be good to have an API that both apps consume for this kind of look up data, including translations there i guess.

jsugarman (approved):
One comment that may be worth talking about at a dev session generally but otherwise looks like a good start 👍

config/locales/en/prior_authority.yml
patrick-laa: I've added a ticket to Trello to discuss, cos you're right. I think in NSM they include translated values as part of the payload, but I think I remember David saying he regretted that approach.

PR #302, opened 2024-02-23
CI: Add tests to merged pr flow

jsugarman:
## Description of change
Add tests to merged pr flow

Inline with agreed changes from laa-submit-crime-forms
repo changes to CI.

see [PR for laa-submit-crime-forms](https://github.com/ministryofjustice/laa-submit-crime-forms/pull/575)
for details


patrick-laa:
Maybe this is Friday brain, but it looks to me like when this runs (not on main), it builds the docker image in `build-opened-pr`->`build-to-ecr`, and we _also_ build it in `test-opened-pr`->`scan-docker-image`. Is there a way we could scan the `build-to-ecr` docker image and save ourselves ~40 seconds?

ivanELEC:
> Maybe this is Friday brain, but it looks to me like when this runs (not on main), it builds the docker image in `build-opened-pr`->`build-to-ecr`, and we _also_ build it in `test-opened-pr`->`scan-docker-image`. Is there a way we could scan the `build-to-ecr` docker image and save ourselves ~40 seconds?

I think this wouldn't work because the executor is different and we need it to be different to set the correct env variables for testing and connect to the right db etc

ivanELEC:
> > Maybe this is Friday brain, but it looks to me like when this runs (not on main), it builds the docker image in `build-opened-pr`->`build-to-ecr`, and we _also_ build it in `test-opened-pr`->`scan-docker-image`. Is there a way we could scan the `build-to-ecr` docker image and save ourselves ~40 seconds?
> 
> I think this wouldn't work because the executor is different and we need it to be different to set the correct env variables for testing and connect to the right db etc

Oh actually I see where you're coming from, we could scan in build-opened-pr only

ivanELEC:
> > > Maybe this is Friday brain, but it looks to me like when this runs (not on main), it builds the docker image in `build-opened-pr`->`build-to-ecr`, and we _also_ build it in `test-opened-pr`->`scan-docker-image`. Is there a way we could scan the `build-to-ecr` docker image and save ourselves ~40 seconds?
> > 
> > 
> > I think this wouldn't work because the executor is different and we need it to be different to set the correct env variables for testing and connect to the right db etc
> 
> Oh actually I see where you're coming from, we could scan in build-opened-pr only

this kind of changes the workflow from build to scan-and-build semantically but I can't see why it wouldn't work

jsugarman:
> Maybe this is Friday brain, but it looks to me like when this runs (not on main), it builds the docker image in `build-opened-pr`->`build-to-ecr`, and we _also_ build it in `test-opened-pr`->`scan-docker-image`. Is there a way we could scan the `build-to-ecr` docker image and save ourselves ~40 seconds?

Ultimately this change does not change the NON-main flow regardless of the merits of this argument or not. I have only changed the name in an attempt to clarify the purpose of the workflows and inline with the name changes on the provider app.

The "old" flow for NOT main
Screenshot 2024-02-23 at 16 59 37

The "new" flow for NOT main
Screenshot 2024-02-23 at 17 01 37

On snyk docker image scanning itself I would be tempted to remove it from circle and use this resuable GHA
https://github.com/ministryofjustice/laa-reusable-github-actions/blob/main/.github/workflows/snyk.yml

patrick-laa (approved):

PR #307, opened 2024-02-26
CRM457-1169: Add all the cards for the overview screen

patrick-laa:
## Description of change
Create a series of Card view models, each of which handles the content for a summary_list_card on the overview page.

I'd suggest we stay flexible about this arrangement of code until we're sure we know when and where these cards may be reused, and with what variations, within the UI.

## Link to relevant ticket

[CRM457-1169](https://dsdmoj.atlassian.net/browse/CRM457-1169)

Screenshot 2024-02-26 at 16 01 44

[CRM457-1169]: https://dsdmoj.atlassian.net/browse/CRM457-1169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/prior_authority/applications/show.html.erb
```diff
@@ -8,13 +8,33 @@
...
-    <%= govuk_summary_card(title: t('.application_details')) { _1.with_summary_list(rows: @details.overview) } %>
+    <%= render 'overview_card', model: @details.overview_card %>
 
     <h2 class="govuk-heading-m"><%= t('.about_the_request') %></h2>
-    <%= govuk_summary_card(title: t('.primary_quote')) do |card| %>
-      <%= govuk_summary_list(rows: @details.primary_quote_rows) %>
+    <%= render 'primary_quote_card', model: @details.primary_quote_card %>
+    <%= render 'reason_why_card', model: @details.reason_why_card %>
+    <% @details.alternative_quotes.each_with_index do |alternative_quote, index| %>
+      <%= render 'alternative_quote_card', model: @details.alternative_quote_card(alternative_quote), n: index + 1 %>
```
dwhenry, line 18: not 100% sure I see the advantage of this pattern over just having teh `alternative_quotes` method return the class??

app/view_models/prior_authority/v1/application_details.rb
```diff
@@ -3,32 +3,59 @@ module V1
...
       end
 
-      private
+      def alternative_quotes
+        quotes.reject { _1['primary'] }.map { Quote.new(_1.merge(item_type:)) }
+      end
+
+      def alternative_quote_card(alternative_quote)
+        AlternativeQuoteCard.new(self, alternative_quote)
+      end
```
dwhenry, line 39: as per comment in view - don;'t see the advantage of this being a separate function..

patrick-laa: Ah, see, what you miss is that the advantage is that this code didn't require me to be thinking clearly to _write_ it...

Yes, you're entirely right. Will fix.

jsugarman (approved):
LGTM

spec/system/prior_authority/view_applications_spec.rb
```diff
@@ -23,15 +23,17 @@
...
+          build(
+            :primary_quote,
+            cost_type: 'per_hour',
+            period: 180,
+            cost_per_hour: '3.50',
+            travel_time: nil,
+            travel_cost_per_hour: nil
+          ),
+          build(:alternative_quote)
+        ]
```
jsugarman, line 36: nit. prefer setup to be outside the example block

spec/system/prior_authority/view_applications_spec.rb
```diff
@@ -43,5 +45,7 @@
     click_on 'LAA-1234'
     expect(page).to have_current_path prior_authority_application_path(application)
     expect(page).to have_content 'Requested: £80.50'
+    click_on 'test.pdf'
```
jsugarman, line 48: should this test not be more extensively testing the expected content of the render "cards" if we are to rely on system tests more than unit tests. coverage will be green but are we covering sufficient permutations and expected content?

jsugarman (approved):
looks good, ✅ , but do have some concerns over testing. nonetheless be good to get this in so we can extend.

spec/system/prior_authority/view_applications_spec.rb
patrick-laa: Yup, will add some more specific expectations today. This ticket is now in QA so I expect there'll be some back and forth before it's merged, which gives me plenty of time to improve the tests.

PR #317, opened 2024-02-28
CRM457 1115: Add quote adjustment tab

jsugarman:
## Description of change
Add quote adjustment elements

[relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1115)

- add key information card
- add service cost section
- add travel cost section
- add additional cost sections

### After changes:

Screenshot 2024-02-28 at 13 44 15

### DIFF
--- a/app/view_models/prior_authority/v1/quote.rb
+++ b/app/view_models/prior_authority/v1/quote.rb
@@ -53,8 +53,24 @@ module PriorAuthority
         NumberTo.pounds(travel_cost_per_hour)
       end
 
+      def formatted_travel_cost_per_unit
+        I18n.t(
+          'per_hour',
+          gbp: travel_cost_per_unit,
+          scope: 'prior_authority.application_details.items.per_unit_descriptions'
+        )
+      end
+
       def base_cost_per_unit
         NumberTo.pounds(cost_type == 'per_item' ? cost_per_item : cost_per_hour)
+      end
+
+      def formatted_base_cost_per_unit
+        I18n.t(
+          cost_type,
+          gbp: base_cost_per_unit,
+          scope: 'prior_authority.application_details.items.per_unit_descriptions'
+        )
       end
 
       def formatted_base_cost
--- a/config/locales/en/prior_authority/applications.yml
+++ b/config/locales/en/prior_authority/applications.yml
@@ -1,38 +1,5 @@
 en:
   prior_authority:
-    adjustments:
-      index:
-        adjust_quote: Adjust quote
-      key_information_card:
-        key_information: Key information
-      additional_costs:
-        additional_cost: Additional cost %{n}
-        adjust_additional_cost: Adjust additional cost
-        adjusted: Caseworker adjusted
-        cost: Cost
-        cost_description: "Cost description: \"%{name} - %{description}\""
-        details: Details
-        requested: Requested
-        time: Time
-        total: Total
-      service_costs:
-        adjusted: Caseworker adjusted
-        adjust_service_cost: Adjust service costs
-        cost: Cost
-        details: Details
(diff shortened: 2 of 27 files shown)


spec/view_models/prior_authority/v1/key_information_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+      expect(key_information.main_offence).to eq('Bad stuff')
+    end
+  end
+
+  describe '#key_information_card' do
+    subject(:key_information) { described_class.new(submission:) }
+
+    let(:submission) { build(:prior_authority_application) }
+
+    it 'returns the postcode from the qute marked as primary true' do
```
jsugarman, line 17:
```suggestion
    it 'returns the key information card object' do
```

app/view_models/prior_authority/v1/quote.rb
```diff
@@ -53,10 +53,26 @@ def travel_cost_per_unit
...
+          gbp: travel_cost_per_unit,
+          scope: 'prior_authority.application_details.items.per_unit_descriptions'
+        )
+      end
+
       def base_cost_per_unit
         NumberTo.pounds(cost_type == 'per_item' ? cost_per_item : cost_per_hour)
       end
 
+      def formatted_base_cost_per_unit
```
patrick-laa, line 68: If the `quote#item_type` is not `"item"` but instead something like `"word"`, then I believe this method should return something like "£10.00 per word" instead of "£10.00 per item"

config/locales/en/prior_authority/applications.yml
```diff
@@ -1,5 +1,38 @@
 en:
   prior_authority:
+    adjustments:
```
patrick-laa, line 3: I think this stuff maybe ought to be in a new file `config/locales/en/prior_authority/adjustments.yml`. At some point (possibly a dev time you missed) we discussed having a larger number of smaller yml files rather than one big monster one.

patrick-laa (approved):
Looking good.

app/view_models/prior_authority/v1/quote.rb
jsugarman: i can get this working in a smelly way but without a matching `cost_per...` method we  fall back to using `cost_per_hour` at line#65 for the value so you would get "£#{cost_per_hour} per word"


```ruby
def formatted_base_cost_per_unit
  scope = 'prior_authority.application_details.items.per_unit_descriptions'

  if I18n.exists?(cost_type, scope:)
    I18n.t(cost_type, gbp: base_cost_per_unit, unit_description: cost_type, scope:)
  else
    I18n.t('other', gbp: base_cost_per_unit, unit_description: cost_type, scope:)
  end
end
```

```yml
per_unit_descriptions:
  per_item: "%{gbp} per item"
  per_hour: "%{gbp} per hour"
  other: "%{gbp} per %{unit_description}"
```

unless we have a `cost_per_other` capability to fallback on and whose key/value is being sent in the JSON it feels dubious. Could get more "meta" or extract to a separate class to handle the logic it but it feels like handling an edge case that would break in other ways at the moment.

what do you think?

config/locales/en/prior_authority/applications.yml
jsugarman: done

app/view_models/prior_authority/v1/quote.rb
patrick-laa: So don't worry, it's not as complicated as you think. `cost_type` will only _ever_ be `per_item` or `per_hour`. It's just that if cost_type is `per_item`, we need to refer to `item_type`, which will be one of `item`, `minute`, `word` or `page`.

PR #320, opened 2024-03-01
CRM457-1151: Preserve original values in payload

patrick-laa:
## Description of change
The first time we make an adjustment to field x, save a `x_original` value to the data blob.

Make it accessible via a new `adjustable_attribute` macro on view models that defines `x` and `x_original` attributes and a convenience `original_x` method that looks for an explicit original and falls back to the non-original.

Also on every adjustment, save the comment into the blob. That way the view models don't need to have anything to do with the events, which become a standalone papertrail.

Add logic into PA that allows it to dynamically calculate both current totals and original totals.

Tidy up NSM calculation logic to use new mechanism

## Link to relevant ticket

[CRM457-1151](https://dsdmoj.atlassian.net/browse/CRM457-1151)

## Screenshot
Screenshot 2024-03-04 at 11 24 20

## Notes for reviewer
The main places to look are:
* app/forms/base_adjustment_form.rb - where we put new stuff into the data blob
* app/view_models/base_view_model.rb - stop stuffing events into view models as we don't need them
* app/view_models/base_with_adjustments.rb - define the "adjustable_attribute" macro to allow easy access to original values
* app/view_models/prior_authority/v1/additional_cost.rb - example of where we use the macro to be able to calculate totals

[CRM457-1151]: https://dsdmoj.atlassian.net/browse/CRM457-1151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/forms/base_adjustment_form.rb
```diff
@@ -29,10 +29,21 @@ def process_field(value:, field:)
...
 
-    selected_record[field] = value
+    ensure_original_field_value_set(field)
+    assign_new_attributes(field, value)
+
     Event::Edit.build(submission:, details:, linked:, current_user:)
   end
 
+  def ensure_original_field_value_set(field)
+    selected_record["#{field}_original"] = selected_record[field] if selected_record["#{field}_original"].nil?
```
dwhenry, line 39: any reason to not just use
```
selected_record["#{field}_original"] ||= selected_record[field]
```

app/view_models/base_with_adjustments.rb
```diff
@@ -1,19 +1,15 @@
...
-    return unless field
-
-    field.details['comment']
+  attribute :adjustment_comment
+
+  def self.adjustable_attribute(attribute_name, type, options = {})
+    attribute attribute_name, type, **options
+    attribute :"#{attribute_name}_original", type, **options
+    define_method :"original_#{attribute_name}" do
+      attributes["#{attribute_name}_original"] || attributes[attribute_name.to_s]
```
dwhenry, line 8: I don't think I like this... but that is because I think the naming will become confusing..

don't have any better suggestions - but will think on this. :)

dwhenry (approved):
I feel significantly upset that this is such a better solution... :(

dwhenry:
Just to confirm on `adjustment_comment` is it also now being stored in the data hash? if so should we use that is the `any_adjustment?` check (as an adjustment must have a comment)

PR #321, opened 2024-03-01
CRM457-1114: Add forms for adjusting quotes

jsugarman:
## Description of change
Add the service cost and travel cost forms forms to enable caseworker adjustments of quotes

[link to ticket](https://dsdmoj.atlassian.net/browse/CRM457-1114)

- update adjust quotes index page to use new storage for original and adjusted values
- add forms to amend
- add javascript for calculation buttons

Note this has not dealt with the additional costs, except for some formatting/content changes. I will pick that up on top of these changes to avoid the branch going stale.

https://www.loom.com/share/e989661cb9f0471ab77da6814f39114b?sid=863c83b9-4eb7-48c8-8ac2-a31fe1ac66e4


app/attributes/type/gbp.rb
```diff
@@ -0,0 +1,19 @@
+module Type
+  class Gbp < ActiveModel::Type::Decimal
+    def cast(value)
+      if suitable_for_casting?(value)
+        # When a monetary value is provided by a form submission, we are only interested
+        # in the value to 2 decimal places, and discard any more fine detail immediately
+        super(value&.to_s&.delete(',')&.delete('£'))&.round(2)
```
dwhenry, line 7: I don't like this as the following would be valid:
```
',23,23£23'
```

jsugarman: This is straight out of provider app! but will take a look

app/view_models/prior_authority/service_costs_with_adjustments.rb
```diff
@@ -0,0 +1,138 @@
...
+    def adjusted_cost_per_item
+      cost_per_item if cost_per_item_original || items_original
+    end
+
+    def adjusted_cost_per_hour
+      cost_per_hour if cost_per_hour_original || period_original
+    end
+  end
+
+  module ServiceCostsWithAdjustments
```
dwhenry, line 134: not sure I understand the benefit of splitting this into multiple modules?

dwhenry (commented):
It's long... :) will have another look tomorrow and likely approve then

app/view_models/prior_authority/service_costs_with_adjustments.rb
jsugarman: this came out of having one monster of a module, originally, violating rubocop module metrics, and me iterating on breaking it down. I could separate the modules, for this and travel costs, making 4 modules, and add them to the quote class individually or leave like this. what you think?

app/attributes/type/gbp.rb
patrick-laa: Yeah, this came from CCQ. Our reasoning was that the problem we were trying to solve was (a) accepting and parsing reasonably-formatted values, such as ones with commas and pound signs, while (b) offering fast feedback if the user has typo'd or misunderstood the question, and not silently converting what the user entered into something they didn't intend. We're not trying to _restrict_ users with our validations, we're trying to _help_ them. So given that no one is going to enter `,23,23£23` via typo or misunderstanding the question, there's not a big benefit to validating a case like that, but there _is_ a cost, namely the cost of me having to write a regex, which I'm not very good at...

app/javascript/component/prior_authority/service_cost_adjustment.js
```diff
@@ -0,0 +1,84 @@
+function init() {
+  const hoursField = document.getElementById('prior_authority_service_cost_form_period_1i');
+  const minutesField = document.getElementById('prior_authority_service_cost_form_period_2i')
+  const costPerHourField = document.getElementById('prior-authority-service-cost-form-cost-per-hour-field')
+  const itemsField = document.getElementById('prior-authority-service-cost-form-items-field')
+  const costPerItemField = document.getElementById('prior-authority-service-cost-form-cost-per-item-field')
+  const calculateChangeButton = document.getElementById('calculate_change_button');
+  const AdjustedCost = document.getElementById('adjusted-cost');
```
jsugarman, line 8:
```suggestion
  const adjustedCost = document.getElementById('adjusted-cost');
```

app/javascript/component/prior_authority/travel_cost_adjustment.js
```diff
@@ -0,0 +1,56 @@
+function init() {
+  const hoursField = document.getElementById('prior_authority_travel_cost_form_travel_time_1i');
+  const minutesField = document.getElementById('prior_authority_travel_cost_form_travel_time_2i')
+  const costPerHourField = document.getElementById('prior-authority-travel-cost-form-travel-cost-per-hour-field')
+  const calculateChangeButton = document.getElementById('calculate_change_button');
+  const AdjustedCost = document.getElementById('adjusted-cost');
```
jsugarman, line 6:
```suggestion
  const adjustedCost = document.getElementById('adjusted-cost');
```

spec/system/prior_authority/adjust_travel_costs_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+    expect(page).to have_content('There are no changes to save. ' \
+                                 'Select cancel if you do not want to make any changes.')
+  end
+
+  it 'allows me to cancel and return to adjustments page' do
+    click_on 'Cancel'
+    expect(page).to have_title('Adjustments')
+  end
+
+  context 'when I calculate the adjusted values', :javascript do
```
jsugarman, line 94: remove the context

app/view_models/prior_authority/service_costs_with_adjustments.rb
dwhenry: This is fine... might be worth adding as a comment so people understand why it's like this... :)

app/view_models/prior_authority/service_costs_with_adjustments.rb
```diff
@@ -0,0 +1,139 @@
...
+        "#{I18n.t(i18n_key, scope: 'prior_authority.application_details.items.per_unit_descriptions')}"
+    end
+
+    def requested_formatted_service_cost_total
+      NumberTo.pounds(requested_base_cost)
+    end
+
+    private
+
+    def requested_items
```
patrick-laa, line 25: There are a few of these. Might be neater to use `alias_attribute`/`alias_method`. But equally, do we really need to create all these additional names for existing methods?

app/view_models/prior_authority/service_costs_with_adjustments.rb
```diff
@@ -0,0 +1,139 @@
...
+    def adjusted_hour_cost_total
+      return unless adjusted_period && adjusted_cost_per_hour
+
+      (
+        (adjusted_period.hours * adjusted_cost_per_hour) +
+        ((adjusted_period.minutes / 60.0) * adjusted_cost_per_hour)
+      ).round(2)
+    end
+
+    def adjusted_items
```
patrick-laa, line 117: The logic of these next few methods is a bit unintuitive to me. Because "adjusted_items" will return a value if "items" hasn't been adjusted (if cost_per_item _has_ been adjusted). This method is kinda trying to do too much. I think it would make more sense to me to separate out a method called `any_per_item_adjustments?` that would check for `items_original` and `cost_per_item_original`. The places that currently call `adjusted_items` in a guard clause could instead call `any_per_item_adjustment?`. And the other places that currently call `adjusted_items` could then just call `items` instead.

patrick-laa (commented):
THe only bits I'm not sure about are the bits you had to rewrite in a hurry after I merged my spike that blew up all the stuff you'd already written to check for adjustments. Sorry again about that...

app/view_models/prior_authority/service_costs_with_adjustments.rb
jsugarman: refactored as suggested and was able to remove the `adjusted_..` pseudo attribute wrappers, so more succinct and hopefully simpler/clearer.

patrick-laa (approved):

PR #329, opened 2024-03-06
Switch to New Namespaces

ivanELEC:
## Description of change

- Update readme references and links
- Update deploy script for branch builder to point to new chart
- Update database references
- Update helm chart

## Link to relevant ticket

[Setup New Kubernetes Namespaces for Caseworker App](https://dsdmoj.atlassian.net/browse/CRM457-941)


.circleci/config.yml
```diff
@@ -8,8 +8,8 @@ orbs:
   ruby: circleci/ruby@2.1.1
   browser-tools: circleci/browser-tools@1.4.7
   snyk: snyk/snyk@1.1.2
-  aws-cli: circleci/aws-cli@4.1.3
-  aws-ecr: circleci/aws-ecr@9.0.2
```
dwhenry, line 12: ??? this reverts a required change to avoid end of life on the linux builds...

.circleci/config.yml
```diff
@@ -299,19 +299,17 @@ jobs:
...
-          extra_build_args: >-
-            --build-arg APP_BUILD_DATE=${APP_BUILD_DATE}
-            --build-arg APP_BUILD_TAG=${CIRCLE_SHA1}
-            --build-arg APP_GIT_COMMIT=${CIRCLE_SHA1}
-            --build-arg APP_BRANCH_NAME=${CIRCLE_BRANCH}
+          extra-build-args: |
+            --build-arg APP_BUILD_DATE=$(date +%Y-%m-%dT%H:%M:%S%z) \
+            --build-arg APP_BUILD_TAG=${CIRCLE_SHA1} \
+            --build-arg APP_GIT_COMMIT=${CIRCLE_SHA1} \
+            --build-arg APP_BRANCH_NAME=${CIRCLE_BRANCH} \
```
dwhenry, line 312: This all supports the above orb updates and shouldn't be reverted

dwhenry (changes requested):
need to fix the oirb stuff.

.circleci/config.yml
ivanELEC: I must have been working from a local version of the file which had the old values

ivanELEC: will look at what's in main and update

dwhenry (approved):

helm_deploy/templates/deployment-worker.yaml
```diff
@@ -151,7 +138,7 @@ spec:
             - name: RAILS_LOG_TO_STDOUT
               value: enabled
             - name: ENV
-              value: {{ .Values.secrets.environment}}
+              value: {{ .Values.variables.environment}}
```
jsugarman, line 141: **nit** variables feels like it could become a catch all config section for our future selves. 

`environment` could be top level and appStoreUrl could be `appStore.Url`

jsugarman: also, i know this already existed, but just to flag that in rails `ENV['RAILS_ENV']` is always `production` on a hosted "environment". The term `host` or `host_env` is more informative to distinguish hosted environments and rails environments IMO.

ivanELEC: yeah I feel like we cross that bridge when we get there in terms of having more environment variables but I agree with your point

helm_deploy/templates/deployment.yaml
```diff
@@ -128,17 +115,17 @@ spec:
             - name: APP_STORE_TENANT_ID
               valueFrom:
                 secretKeyRef:
-                  name: azure-secret
+                  name: app-store-auth
```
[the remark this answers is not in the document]
ivanELEC: yeah they match :)

PR #334, opened 2024-03-08
CRM457-1114: update overview to show original values

jsugarman:
## Description of change
Update overview tab to display original/requested values

[link to ticket](https://dsdmoj.atlassian.net/browse/CRM457-1220)
[link to parent ticket](https://dsdmoj.atlassian.net/browse/CRM457-1114)

Following the adjusted quotes work the overview page needs to 
ensure it shows the original provider supplied quote values
for the primary quote

### DIFF
--- a/app/view_models/prior_authority/v1/additional_cost.rb
+++ b/app/view_models/prior_authority/v1/additional_cost.rb
@@ -33,7 +33,7 @@ module PriorAuthority
       end
 
       def formatted_total_cost
-        NumberTo.pounds(total_cost)
+        NumberTo.pounds(total_cost(original: true))
       end
 
       def form_attributes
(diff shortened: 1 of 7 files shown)


dwhenry (approved):

app/view_models/prior_authority/v1/additional_cost.rb
```diff
@@ -33,7 +33,7 @@ def total_item_cost(original)
       end
 
       def formatted_total_cost
-        NumberTo.pounds(total_cost)
+        NumberTo.pounds(total_cost(original: true))
```
patrick-laa, line 36: Does it explicitly say somewhere that on the overview page we should show original values rather than adjusted values? I couldn't see it on a skim of the tickets but might have missed it. It just seems a little bit unintuitive UX to me.

jsugarman: I queried this with UCD lot and they said it should show the original/requested values when adjustments exist. The prototype does too, although i queried because i was not sure that was intentional

patrick-laa: Ok nice one, you're on step ahead of me. Then this all looks fine to me and I think David has already approved.

PR #336, opened 2024-03-08
Extract coverage to separate job and collate

jsugarman:
## Description of change
Extract coverage to separate job and collate results

Allow parallel test suite runs and separate test failures 
from coverage failures.

Also remove codeclimate because:
a. we have sonarcloud wired up which does a lot of the duplication and quality analysis already, 
b.  rubocop is already linting as part of the test suite,
c.  it clutters the circleci config with setup for it.
d. its never been in, or currently isn't in, provider app.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,21 +106,6 @@ commands:
     steps:
       - install-gems
       - install-compile-assets
-      - install-codeclimate-tt
-
-  install-codeclimate-tt:
-      steps:
-        - run:
-            name: Install codeclimate test reporter
-            command: |
-              mkdir -p tmp/
-              mkdir -p tmp/coverage/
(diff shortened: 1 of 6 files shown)


spec/system/feedback_spec.rb
```diff
@@ -5,6 +5,7 @@
     sign_in create(:caseworker)
     visit root_path
     click_on 'Feedback'
+    ActiveJob::Base.queue_adapter.enqueued_jobs.clear
```
patrick-laa, line 8: Nit: this feels like something that, if it's needed, should be handled by a `config.after(:each, type: :system) { ... }` block in the rails_helper file so that we don't need to remember to do it in every system spec where it matters.

patrick-laa (approved):

PR #340, opened 2024-03-11
[WIP] Crm457 984/auto grant

dwhenry:
## Description of change
Autogrant new application as tehy arrive

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-984)


README.md
```diff
@@ -31,6 +31,9 @@ DATABASE_URL=postgresql://john.smith@localhost/laa-claim-non-standard-magistrate
 
 After you've defined your DB configuration in the above files, run the following:
 
+> NOTE: you need to manually instal postgis for the migrations to run. This can be done with
```
patrick-laa, line 34: I don't think this is true - postgis is included in the brewfile now

.env.development
```diff
@@ -24,9 +24,9 @@ SIDEKIQ_WEB_UI_PASSWORD=sidekiq
 # App store credentials (for authenticating to a local app store currently)
 # use dev app-store-auth creds.
 # TODO: OAuth requirment may be removed from local/development app store in the future
```
patrick-laa, line 26: As this todo has now happened, we should just remove these vars as they are not needed and all they will do is break stuff

app/services/autograntable.rb
```diff
@@ -0,0 +1,84 @@
+class Autograntable
```
patrick-laa, line 1: I really like this code. Really clear, nicely divided up into sensible chunks. Big thumbs up from me.

app/jobs/pull_latest_version_data.rb
```diff
@@ -29,11 +29,37 @@ def save(submission, json_data)
...
+
+      autogrant(submission) if cached_autograntable
+    end
+  end
+
+  def autograntable(submission:)
+    # performed here to avoid slow transactions as requires API call to the OS API
+    Autograntable.new(submission:).grantable?
+  rescue LocationService::LocationError
+    false
```
patrick-laa, line 46: If the location error is anything other than a LocationService::NotFoundError then that suggests that something has happened that we may need to react to, like an API key having expired. Should we do a Sentry notification for that as part of the handling?

spec/factories/prior_authority_applications.rb
```diff
@@ -89,6 +90,20 @@
...
+    trait :per_hour do
+      cost_type { 'per_hour' }
+      cost_per_hour { '80.0' }
+      cost_per_item { nil }
+      item_type { nil }
+      period { 420 }
+    end
+
+    trait :no_travel do
+      travel_time { 0 }
```
patrick-laa, line 103: Because the travel detail form is skipped by default unless the user clicks through to it from the summary screen, I think travel_time should never be 0 - either the screen is skipped, so it's nil, or the form is filled out, in which case validations force it to be > 0. So I think this in unrealistic data.

patrick-laa (approved):
Love the core code. A few little comments about the stuff around it.

README.md
dwhenry: might be my version of postgres... I will check I'm on the latest

spec/factories/prior_authority_applications.rb
dwhenry: 👍 will update

PR #342, opened 2024-03-12
Email Notification Start PR

ivanELEC:
## Description of change

 Namespace CRM7 email notifications so we can do CRM4  ones

## Notes for reviewer
This is not a full PR and doesn't have 100% coverage as it's missing else pathways (this is where CRM4 things would go so just should wait til that's tested) - can be used as a basis for any of the email decision notification tickets


jsugarman:
closed as superseded

PR #348, opened 2024-03-13
CRM457-1218: Add location service to check if postcode inside M25

patrick-laa:
## Description of change
- Generated a KML file plotting the M25 from a route generated via the OSRM API
- Added a call to the OS API to convert a postcode into some  OSGB coordinates
- Use postgis to convert OSGB into ~sane~ more standard coordinates and check whether the point falls inside the polygon defined by the KML

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1218)

You can test for yourself using an API key generated at https://osdatahub.os.uk/projects


dwhenry (approved):
Would like a comment on how you make the KML file so that if we need to replicate/modify in the future that is clearly documented.

jsugarman (commented):
Be good to have a section of the readme to cover this, and the setup needed locally.

PR #353, opened 2024-03-15
CRM457-976/980: Namespace mailers and wire up prior authority templates

jsugarman:
## Description of change
Namespace mailers and wire up prior authority templates

[Decision notifications](https://dsdmoj.atlassian.net/browse/CRM457-976)
[Further information notification](https://dsdmoj.atlassian.net/browse/CRM457-980)

To differentiate at a low level this splits them to their own directories  and amends the PA mailers so that they:
- call the correct/PA notify templates
- pass the current expected personalisation arguments, including calculated totals

### Screenshots of emails

GRANTED

Screenshot 2024-03-18 at 11 28 27

PART GRANTED

Screenshot 2024-03-18 at 11 29 42

REJECTED

Screenshot 2024-03-18 at 11 30 10


app/controllers/about/feedback_controller.rb
```diff
@@ -19,10 +19,11 @@ def create
     private
 
     def submit_feedback
-      FeedbackMailer.notify(**feedback_params.to_h
-                                             .symbolize_keys
-                                             .merge(application_env: application_environment))
-                    .deliver_later!
+      Nsm::FeedbackMailer
```
patrick-laa, line 22: I don't think this one should be namespaced - this is the generic feedback mailer for the feedback form that is accessible from the footer of all parts of the site, whether NSM or PA.

app/mailers/prior_authority/feedback_messages/feedback_base.rb
```diff
@@ -0,0 +1,51 @@
...
+      def case_reference
+        @submission.data['laa_reference']
+      end
+
+      def ufn
+        @submission.data['ufn']
+      end
+
+      def application_total
+        'TODO'
```
patrick-laa, line 42: I think we've got this, don't we? from ApplicationSummary.formatted_original_total_cost

app/mailers/prior_authority/feedback_messages/part_granted_feedback.rb
```diff
@@ -0,0 +1,28 @@
...
+          application_total: application_total,
+          part_grant_total: adjusted_total,
+          caseworker_decision_explanation: @comment,
+          date: DateTime.now.strftime('%d %B %Y'),
+          feedback_url: feedback_url
+        }
+      end
+
+      def adjusted_total
+        'TODO'
```
patrick-laa, line 24: And this one is ApplicationSummary.formatted_total_cost. Or is there another ticket where we should do that?

spec/factories/prior_authority_applications.rb
```diff
@@ -12,6 +12,12 @@
     initialize_with { attributes }
     laa_reference { 'LAA-123456' }
     prison_law { false }
+    provider do
+      {
+        'email' => 'provider@example.com',
+        'description' => nil, # TODO: is this a thing?
```
patrick-laa, line 18: It looks like the provider app sends over a description, which is a value it's received from Portal. But on UAT the description we've received for the test provider is nil, so it might always be nil. 🤷

app/controllers/about/feedback_controller.rb
jsugarman: good catch, the template is the same and looks generic enough to be shared at the moment so...done.

app/mailers/prior_authority/feedback_messages/part_granted_feedback.rb
jsugarman: thanks for the pointer, I have added that now, with tests

app/mailers/prior_authority/feedback_messages/feedback_base.rb
jsugarman: thanks for the pointer, I have added that now, with tests.

app/services/notify_app_store.rb
```diff
@@ -8,7 +8,9 @@ def self.process(submission:)
     else
       begin
         new.notify(MessageBuilder.new(submission:))
-        SubmissionFeedbackMailer.notify(submission).deliver_later!
+
+        klass = "#{submission.namespace}::SubmissionFeedbackMailer".constantize
```
dwhenry, line 12: just as a note an alternative to `constantize`

```
submission.namespace.const_get('SubmissionFeedbackMailer')
```

app/view_models/prior_authority/v1/decision.rb
```diff
@@ -7,7 +7,7 @@ class Decision < BaseViewModel
       delegate :state, to: :submission
 
       def comments
-        submission.latest_decision_event.details['comment']
+        submission.latest_decision_event&.details&.fetch('comment', '')
```
dwhenry, line 10: is this comment not now stored in the JSON blob?

dwhenry (approved):
LGTM - small comments

app/services/notify_app_store.rb
jsugarman: cool, using this just for cosistency with in `app/view_models/base_view_model.rb`.

app/view_models/prior_authority/v1/decision.rb
jsugarman: there are multiple values in the JSON blob, so could either rely on latest decision event, as above, or combination of current state plus the relevant attribute below logic. currently latest decision event seems the most sensible to me especially as I am not sure whether a further information request is left in a status of submitted for which there is no  "pending..explanation" state.

```
   # payload change when part granting...
   "pending_granted_explanation"=>"",
   "pending_rejected_explanation"=>"",
   "pending_part_grant_explanation"=>"I have part granted this because....",
```

As an aside, should we event be sending an email immediately, if the decision is "pending"?!

patrick-laa: Sorry, this is my fault - the reason these are all prefixed with `pending` is that when I wrote this logic, I was thinking of the 'save and come back later' route - we were saving these comments in the blob purely so that we could re-access them when we came back later, so _in that scenario_ these would be pending values because a decision would not have been made. However, those values also get written when we actually finalise the decision, so they are technically available to us here when we're viewing the post-decision screen. But possibly they _shouldn't_ be there - maybe pending decisions should be cleared out when a decision is actually made, in which case the canonical source of truth for the comment is the latest decision event.

PR #356, opened 2024-03-18
Fix missing value

patrick-laa:
Otherwise the "open application" tab breaks.


patrick-laa:
Closing as addressed in another PR. The issue is with bad data, not bad code.

PR #360, opened 2024-03-18
Crm457 1240/event users

dwhenry:
## Description of change
Fix bug with access table not being visible in branch environments due to missing data - caused by invalid events

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1240)

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -27,11 +27,30 @@ class Event < ApplicationRecord
     private :create
 
     def rehydrate!(params)
+      check_user(params) unless HostEnv.production?
+
       new(params).save!
     end
 
     def latest_decision
       order(created_at: :desc).find_by(event_type: 'Event::Decision')
+    end
+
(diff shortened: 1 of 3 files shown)


app/models/event.rb
```diff
@@ -27,12 +27,31 @@ class << self
...
       new(params).save!
     end
 
     def latest_decision
       order(created_at: :desc).find_by(event_type: 'Event::Decision')
     end
+
+    private
+
+    def check_user(params)
```
patrick-laa, line 41: Given this method is doing something a bit yick (in that we really wouldn't want this happening in production), can we flag it up with a less innocuous method name? `create_dummy_user_if_user_id_not_recognised` is maybe going a bit far, but you get the picture - something to make clear that this is a slightly risky procedure.

patrick-laa (approved):

PR #366, opened 2024-03-19
CRM457-1186: Add resubmission deadline

patrick-laa:
## Description of change
We already are sending almost everything we need to the app store when we send back an RFI. This PR just:
- Adds the resubmission deadline for display in provider UI as per the ticket
- Tidies up the code
- Adds tests to prove the right info is stored.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1186)

### DIFF
--- a/app/forms/prior_authority/send_back_form.rb
+++ b/app/forms/prior_authority/send_back_form.rb
@@ -39,6 +39,8 @@ module PriorAuthority
     end
 
     def stash(add_draft_send_back_event: true)
+      updates_needed.compact_blank! # The checkbox array introduces an empty string value
+
       submission.data.merge!(attributes.except('submission', 'current_user'))
       submission.save!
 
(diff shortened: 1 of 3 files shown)


app/forms/prior_authority/send_back_form.rb
```diff
@@ -39,6 +39,8 @@ def save
     end
 
     def stash(add_draft_send_back_event: true)
+      updates_needed.compact_blank! # The checkbox array introduces an empty string value
```
jsugarman, line 42: **nit** IMO we do not need the comment

spec/forms/prior_authority/send_back_form_spec.rb
```diff
@@ -46,4 +46,73 @@
...
+      let(:params) do
+        {
+          updates_needed: ['further_information'],
+          further_information_explanation: '',
+          submission: submission
+        }
+      end
+
+      it 'returns false' do
+        expect(subject.save).to be false
```
jsugarman, line 70: **nit and outside scope of work**, and i guess we do not have rspec rubocop in or configured, but i like the cop and logic that encourages named subjects - `form.save` is much more readable but either way

jsugarman (approved):

PR #371, opened 2024-03-20
CRM457-1187: Tidy states and UI for showing sent back applications

patrick-laa:
## Description of change
Adjust UI for showing applications to accommodate 'sent_back' state
Moved sent_back applications to the 'Open' table, not the 'Assessed' table

Tried to standardise naming around states. As these have now diverged quite a lot from NSM states, moved state-specific scopes into submission subclasses

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1187)

Screenshot 2024-03-20 at 11 18 36

### DIFF
--- a/spec/mailers/prior_authority/submission_feedback_mailer_spec.rb
+++ b/spec/mailers/prior_authority/submission_feedback_mailer_spec.rb
@@ -159,11 +159,11 @@ RSpec.describe PriorAuthority::SubmissionFeedbackMailer, type: :mailer do
     include_examples 'creates a prior authority feedback mailer'
   end
 
-  context 'with other state for further information request' do
+  context 'with an unhandled state' do
     let(:application) do
       create(
         :prior_authority_application,
-        state: 'fake',
+        state: 'submitted',
         data: build(
           :prior_authority_data,
           laa_reference: 'LAA-FHaMVK',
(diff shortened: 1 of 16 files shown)


spec/mailers/prior_authority/submission_feedback_mailer_spec.rb
```diff
@@ -133,7 +133,7 @@
   end
 
   context 'with other state for further information request' do
-    let(:state) { 'fake' }
+    let(:state) { 'sent_back' }
```
jsugarman, line 136: i think this file needs rebasing as it is not showing latest intent for this test. or the one above at line 119

jsugarman: it should be just testing an invalid state. 

```
context 'with other state' do
```

patrick-laa: @jsugarman the issue is we now have validation that prevents unrecognised states being persisted. But maybe that's a sign this spec is now redundant.

jsugarman: yes, perhaps, although the latest version is just testing for error capture and raising of invalid states. The test at 119 above it, in main, is already exercising the 'sent_back' state.

patrick-laa: Ok, rebased now. It's impossible for a PA application to be assigned a state that's not in its list, but the spec still works if we assign a state that's valid but we don't have a feedback mailer for. Like `submitted`.

jsugarman (approved):

PR #374, opened 2024-03-21
Minimise use of Rails.env

patrick-laa:
## Description of change

Setting up an environment for e2e tests is proving to be a pain because different Rails envs are hard-coded to behave in different ways, none of which quite forms for e2e. This makes everything more env-var-y


imtiazAhmd (approved):
LGTM

jogilder (commented):
@patrick-laa interesting to see a LOCAL_STACK environment variable as well. 
I think it would be useful to document all these settings and what they do. I'll make a note to create a ticket for when we have time.

PR #394, opened 2024-03-28
CRM457-1273: Make layout more consistent and avoid rows inside columns

patrick-laa:
## Description of change

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1273)


dwhenry (approved):
have no idea - but assume everything is better

PR #405, opened 2024-04-05
Do not try to save events we already have records of

patrick-laa:
Events are immutable, so if we already have a record of an event we don't need to update it. But we do need not to try to recreate it, otherwise we end up with duplicates.


jogilder (approved):
LGTM

PR #411, opened 2024-04-10
Add Metabase Image Pod [DO NOT MERGE]

ivanELEC:
## Description of change

 [SPIKE - Consider if we could use metabase in the same way as Crime Apply](https://dsdmoj.atlassian.net/browse/CRM457-1278)

- Used CCQ code as a basis
- Added metabase image to helm deploy
- Added feature flag and enabled metabase in only dev
- Added service host to connect to metabase via URL


jsugarman:
closing as we have decided to use a UAT metabase instance for targeting all pre-prod databases for analytics

PR #412, opened 2024-04-11
CRM457-1327: Pub/Sub instead of cron

patrick-laa:
## Description of change

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1327)


app/jobs/pull_updates.rb
```diff
@@ -4,28 +4,10 @@ class PullUpdates < ApplicationJob
   def perform
     last_update = Submission.maximum(:app_store_updated_at) || Time.zone.local(2023, 1, 1)
 
-    json_data = HttpPuller.new.get_all(last_update)
+    json_data = AppStoreClient.new.get_all_submissions(last_update)
 
     json_data['applications'].each do |record|
-      save(record['application_id'], convert_params(record))
+      ReceiveApplicationMetadata.new(record).save
```
dwhenry, line 10: as per my submit message... I no longer think saving here is the correct idea... this should just create a job to pull all the data in.. :)

dwhenry (approved):
so... hoops because 2 stage saving..

if we consolidate the save to one point that "should" simplify things (IMHO)

have approved as this code is technically fine and could be merged as is.

PR #420, opened 2024-04-18
Crm457 785 core costs

dwhenry:
## Description of change
 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-785)

This is only only part of the requirements on the ticket - the rest will be split out into a new tickets

This specifically implements:
* new core cost view at the top of the adjustments tab
* additional changes that have been strike-though in the list at the bottom

Now implemented:
* Updated Work items summary
* Updated Work items tables
* Anything not striked out in the list at then bottom

### Before changes:
Screenshot 2024-04-18 at 12 10 22
Screenshot 2024-04-18 at 12 10 31

### After changes:
Screenshot 2024-04-18 at 12 10 02
Screenshot 2024-04-18 at 12 09 52

### DIFF
--- a/spec/system/accessibility_spec.rb
+++ b/spec/system/accessibility_spec.rb
@@ -22,7 +22,6 @@ RSpec.describe 'Accessibility', :accessibility do
   context 'when viewing claim-specific screens' do
     %i[nsm_claim_claim_details
        nsm_claim_adjustments
-       nsm_claim_letters_and_calls
        nsm_claim_supporting_evidences
        nsm_claim_history
        edit_nsm_claim_change_risk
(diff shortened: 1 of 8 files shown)


spec/system/accessibility_spec.rb
```diff
@@ -22,7 +22,6 @@
   context 'when viewing claim-specific screens' do
     %i[nsm_claim_claim_details
        nsm_claim_adjustments
-       nsm_claim_letters_and_calls
```
patrick-laa, line 25: Given that you've fixed what this spec was complaining about, I don't think you need to remove this here any more, do you? It seems to pass on my machine when I put this line back in.

patrick-laa (approved):

spec/system/accessibility_spec.rb
dwhenry: it's a duplicate - it was failing in two spots... the grouping is a little off, but descided it was better the way I left it.

PR #423, opened 2024-04-22
CRM457-762: Show totals at top (and bug-fixes)

patrick-laa:
## Description of change
Add "Claimed" "Allowed" totals to the top of NSM claims
Also fix some bugs - explanations in comments.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-762)

### DIFF
--- a/app/mailers/nsm/feedback_messages/feedback_base.rb
+++ b/app/mailers/nsm/feedback_messages/feedback_base.rb
@@ -58,7 +58,7 @@ module Nsm
       end
 
       def claim_total
-        @submission.data['submitted_total_inc_vat'] || @submission.data['submitted_total'] || 0
+        @submission.formatted_claimed_total
       end
     end
   end
--- a/app/view_models/nsm/v1/core_cost_summary.rb
+++ b/app/view_models/nsm/v1/core_cost_summary.rb
@@ -51,7 +51,7 @@ module Nsm
       def calculate_profit_costs(formatted:)
         letters_calls = LettersAndCallsSummary.new('submission' => submission).rows
 
-        calculate_row(work_items[PROFIT_COSTS] || ([] + letters_calls), 'profit_costs', formatted:)
+        calculate_row((work_items[PROFIT_COSTS] || []) + letters_calls, 'profit_costs', formatted:)
       end
 
       def calculate_waiting(formatted:)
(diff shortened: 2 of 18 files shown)


app/mailers/nsm/feedback_messages/feedback_base.rb
```diff
@@ -58,7 +58,7 @@ def ufn
       end
 
       def claim_total
-        @submission.data['submitted_total_inc_vat'] || @submission.data['submitted_total'] || 0
```
patrick-laa, line 61: This field was never populated. I'm replacing it with a formatted string - I don't know if that will be a problem in Notify when this value is interpolated - if so we'll need to rewrite CoreCostSummary to give access to the raw numbers after it does the calculation of totals.

app/view_models/nsm/v1/core_cost_summary.rb
```diff
@@ -51,7 +51,7 @@ def sum_allowed(data, field)
       def calculate_profit_costs(formatted:)
         letters_calls = LettersAndCallsSummary.new('submission' => submission).rows
 
-        calculate_row(work_items[PROFIT_COSTS] || ([] + letters_calls), 'profit_costs', formatted:)
+        calculate_row((work_items[PROFIT_COSTS] || []) + letters_calls, 'profit_costs', formatted:)
```
patrick-laa, line 54: This was a bug preventing letters and calls from  being included in profit costs

app/views/nsm/claims/_claim_summary.html.erb
```diff
@@ -1,17 +1,30 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <% if claim.display_state?  %>
-        <p class="govuk-!-margin-bottom-2">
-          <%= govuk_tag(text: t(".status_list.#{'granted'}"), colour: accessed_colour('granted')) %>
```
patrick-laa, line 5: This was my bad - I hard-coded the state in an earlier PR while testing something out and didn't notice that I forgot to take it out again!

dwhenry (approved):
So one of the reasons we were planning to have the total in the JSON hash was to help with reporting. I still think this would be useful to have in the data set - but maybe just  wait until we have a reporting requirement and add it then. :)

PR #440, opened 2024-04-29
Reduce app store calls

patrick-laa:
## Description of change
Use `params[:data]` instead of re-pulling the data via a GET request
This means we can remove the `get_submission` app store call entirely, as we never use it.

### DIFF
--- a/spec/requests/sync_spec.rb
+++ b/spec/requests/sync_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe 'Syncs' do
 
   describe 'POST /app_store_webhook' do
     let(:client) { instance_double(AppStoreClient) }
-    let(:record) { :record }
+    let(:record) { { 'foo' => 'bar' } }
 
     context 'when no auth token is provided' do
       it 'rejects all requests' do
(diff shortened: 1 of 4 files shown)


spec/requests/sync_spec.rb
```diff
@@ -78,15 +75,12 @@
...
           allow(JWT).to receive(:decode).with('ABC', nil, true, { algorithms: 'RS256', jwks: jwks }).and_return(decoded)
 
-          allow(AppStoreClient).to receive(:new).and_return(client)
-          allow(client).to receive(:get_submission).and_return(record)
           allow(UpdateSubmission).to receive(:call)
         end
 
         it 'triggers a sync' do
-          post '/app_store_webhook', params: { submission_id: '123' }, headers: { 'Authorization' => 'Bearer ABC' }
+          post '/app_store_webhook', params: { submission_id: '123', data: record }, headers: { 'Authorization' => 'Bearer ABC' }
```
dwhenry, line 82: do we want to think about the endpoint now that we have chnaged what it does to some extent???

dwhenry (approved):

PR #441, opened 2024-04-29
Paginate before mapping

patrick-laa:
Otherwise when there are 100,000 items in the database, you load them _all_ just to render 10...

### DIFF
--- a/app/controllers/nsm/claims_controller.rb
+++ b/app/controllers/nsm/claims_controller.rb
@@ -1,9 +1,8 @@
 module Nsm
   class ClaimsController < ApplicationController
     def index
-      claims = Claim.pending_decision
-      claims = claims.map { |claim| BaseViewModel.build(:all_claims, claim) }
-      @pagy, @claims = pagy_array(claims)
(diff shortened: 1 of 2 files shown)


jsugarman (approved):

app/controllers/nsm/claims_controller.rb
```diff
@@ -1,9 +1,8 @@
 module Nsm
   class ClaimsController < ApplicationController
     def index
-      claims = Claim.pending_decision
-      claims = claims.map { |claim| BaseViewModel.build(:all_claims, claim) }
-      @pagy, @claims = pagy_array(claims)
+      @pagy, claims = pagy_array(Claim.pending_decision)
```
jsugarman, line 4: pretty sure this line will still kill it. there are ~48k claims on UAT.

```
irb(main):004> require 'pagy/console'
=> true
include Pagy::Console
=> Object
irb(main):006> @pagy, claims = pagy_array(Claim.pending_decision)
=>
Killed
```

PR #455, opened 2024-05-08
CRM457-1284: Add assessment fakery

patrick-laa:
## Description of change
Group assessable submissions into batches so that we don't have to worry about long-running jobs being interrupted
Randomise between all different outcomes

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1284)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -90,8 +90,6 @@ RSpec/MessageChain:
 RSpec/SpecFilePathFormat:
   Enabled: false
 
-
-
 Naming/VariableNumber:
   Enabled: false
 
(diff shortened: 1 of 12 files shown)


ivanELEC (approved):

.rubocop.yml
```diff
@@ -164,3 +164,7 @@ Rails/ActionControllerFlashBeforeRender:
 
 RSpec/FilePath:
   Enabled: false
+
+# Occasionally it's useful to create big lists
+FactoryBot/ExcessiveCreateList:
+  Enabled: false
```
jsugarman, line 170: **nit** - might be worth an explicit `Exlude` for the individual files here if possible

PR #468, opened 2024-05-14
CRM457-1461: Show in-progress tags on "your claims" page

patrick-laa:
## Description of change

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1461)


ivanELEC (approved):
Nit: we could bring in the tag colours like we do in prior authority by having the helper but this is also fine (conscious we have a lot to do so don't want to hold this pr back)

PR #477, opened 2024-05-15
CRM457-1467: Fix floating point rounding errors

patrick-laa:
## Description of change
We were explicitly casting some values as floats, and IntegerTimePeriod multiplications were being quietly converted to floats even when multiplied with decimals. This was leading to rounding errors.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1467)


jsugarman (approved):
Is there any javascript on the page that might need to account for these rounding errors too? i.e. update calculation button on the page

PR #478, opened 2024-05-16
CRM457-1482: Add full sorting to open claims

patrick-laa:
## Description of change
Reuse PA sorting mechanism for open claims

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1482)


app/controllers/nsm/claims_controller.rb
```diff
@@ -32,5 +33,17 @@ def new
...
+    private
+
+    def order_and_paginate(query)
+      pagy(Sorter.call(query, @sort_by, @sort_direction))
+    end
+
+    def set_default_table_sort_options
+      default = 'date_updated'
+      @sort_by = params.fetch(:sort_by, default)
+      @sort_direction = params.fetch(:sort_direction, 'descending')
```
dwhenry, line 46: sorry to be difficult... but surely the default direction got any non-date field is ascending... so..
```
    DEFAULT_SORT_FIELD = 'date_updated'
    def set_default_table_sort_options
      @sort_by = params.fetch(:sort_by, DEFAULT_SORT_FIELD)
      @sort_direction = params.fetch(:sort_direction, @sort_by == DEFAULT_SORT_FIELD ? 'descending' : 'ascending')
    end
```

app/services/sorter.rb
```diff
@@ -0,0 +1,34 @@
...
+         ELSE 'in_progress' END
+       WHEN state = 'further_info' THEN 'sent_back'
+       ELSE state END ?
+  SQL
+
+  ORDERS = {
+    'laa_reference' => "data->>'laa_reference' ?",
+    'firm_name' => "data->'firm_office'->>'name' ?",
+    'client_name' => "(data->'defendant'->>'first_name') || ' ' || (data->'defendant'->>'last_name') ?",
+    'caseworker' => "COALESCE(users.first_name, 'Not') || ' ' || COALESCE(users.last_name, 'assigned') ?",
```
dwhenry, line 16: so I would have assumed you would want not assigned to appear eitehr first or last not in the middle...

app/services/sorter.rb
```diff
@@ -0,0 +1,34 @@
...
+         CASE WHEN users.id IS NULL THEN  'not_assigned'
+         ELSE 'in_progress' END
+       WHEN state = 'further_info' THEN 'sent_back'
+       ELSE state END ?
+  SQL
+
+  ORDERS = {
+    'laa_reference' => "data->>'laa_reference' ?",
+    'firm_name' => "data->'firm_office'->>'name' ?",
+    'client_name' => "(data->'defendant'->>'first_name') || ' ' || (data->'defendant'->>'last_name') ?",
```
dwhenry, line 15: so NSM records can have multiple defendants... :( so not 100% sure this works)

dwhenry (commented):
think there is an issue with defendant sorting as it should be an array of defendants. :(

app/controllers/nsm/claims_controller.rb
patrick-laa: I think we're ok on this - due to how the mechanism works, any time `sort_by` is specified, `sort_direction` is also specified, so the only  time we rely on the default sort direction is when we're also relying on the default sort field, which is a date.

app/services/sorter.rb
patrick-laa: Ah crap. I missed that. Thank you!

patrick-laa: Now fixed

app/services/sorter.rb
```diff
@@ -0,0 +1,44 @@
...
+      base_query.then { add_joins(_1, sort_by) }.order(Arel.sql(order_template.gsub('?', direction)))
+    end
+
+    def add_joins(query, sort_by)
+      with_users = query.left_joins(assignments: :user)
+
+      return with_users unless sort_by == 'main_defendant_name'
+
+      with_users.joins("CROSS JOIN JSONB_ARRAY_ELEMENTS(data->'defendants') as defendants")
+                .where("defendants.value->>'main' = 'true'")
```
dwhenry, line 41: I have no idea what this does... seems complicated.. :)

dwhenry (approved):
LGTM

PR #485, opened 2024-05-21
CRM457-1379: Full disbursement adjustments

patrick-laa:
## Description of change
Depending on if this disbursement is mileage-based, either allow editing of miles or direct editing of the pre-vat total. Also allow editing of whether VAT is enabled or not.

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1379)


config/locales/en/nsm/disbursements.yml
```diff
@@ -78,8 +94,17 @@ en:
         nsm/disbursements_form:
           attributes:
             base:
+              no_change: The new values cannot be the same as the values submitted by the provider.
             total_cost_without_vat:
```
ivanELEC, line 98: Potentially misleading error message/the error is potentially triggered at the wrong time: 

If I get a claim with disbursements, then adjust a disbursement and save, then go back to the disbursement and not change anything (keeping my original adjustment in) - we still get this same error even though these aren't the same values submitted by provider

ivanELEC:
NIT (but I could be convinced it could be more serious): this introduces inconsistencies in how decimals are stored in the payload (some are strings and some are numbers) - could this have a knock on effect for anything that is consuming this data via the API?

image

config/locales/en/nsm/disbursements.yml
patrick-laa: Ok, the `data_has_changed?` method now checks for changes compared to the original values, not the most recently adjusted values.

ivanELEC:
> NIT (but I could be convinced it could be more serious): this introduces inconsistencies in how decimals are stored in the payload (some are strings and some are numbers) - could this have a knock on effect for anything that is consuming this data via the API?
> 
> image

the BaseAdjustmentForm#ensure_original_field_value_set is still setting the "miles_original" field to a string type. Though not exactly sure how to fix this other than extending the method to be able to take in an optional alternate value (but I don't know about this as it undermines the safety of the method always setting to the original value) or extending it to optionally use send to perform additional formatting

ivanELEC (approved):
I'm going to approve this as the only remaining nit doesn't cause any issues with our current build and it can be addressed separately

ivanELEC:
> > NIT (but I could be convinced it could be more serious): this introduces inconsistencies in how decimals are stored in the payload (some are strings and some are numbers) - could this have a knock on effect for anything that is consuming this data via the API?
> > image
> 
> the BaseAdjustmentForm#ensure_original_field_value_set is still setting the "miles_original" field to a string type. Though not exactly sure how to fix this other than extending the method to be able to take in an optional alternate value (but I don't know about this as it undermines the safety of the method always setting to the original value) or extending it to optionally use send to perform additional formatting

Ignore that last - comment, I was looking at a claim that was adjusted and sent back before the changes were made. This works fine

PR #492, opened 2024-05-24
Bump braces 3.0.2 to 3.0.3

jsugarman:
## Description of change
Bump braces 3.0.2 to 3.0.3 and unignore CVE warning

This fixes the CVE we currently ignore, designated
as SNYK-JS-BRACES-6838727.

## Notes for reviewer

If Snyk scan passes when the specific CVE is unignored then it is good to go. The snyk ignore has been left in as an example for future reference.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,10 +1,11 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
 version: v1.25.0
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
-ignore:
-  SNYK-JS-BRACES-6838727:
-    - '*':
-        reason: no fix; no impact
-        expires: 2099-01-01T00:00:00.000Z
-        created: 2024-05-14T06:44:55.382Z
+ignore: {}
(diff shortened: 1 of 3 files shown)


.snyk
```diff
@@ -1,10 +1,11 @@
...
-        reason: no fix; no impact
-        expires: 2099-01-01T00:00:00.000Z
-        created: 2024-05-14T06:44:55.382Z
+ignore: {}
+  # FIXED by bumping to braces@3.0.3
+  # SNYK-JS-BRACES-6838727:
+  #   - '*':
+  #       reason: no fix; no impact
+  #       expires: 2099-01-01T00:00:00.000Z
+  #       created: 2024-05-14T06:44:55.382Z
```
dwhenry, line 10: do we need this here as a comment?

dwhenry (approved):

.snyk
jsugarman: i left this in on submit to as a reference for future selves as it can be a pain to download the snyk cli and generate the file. for consistency with submit alone i would leave in for that reason

PR #502, opened 2024-05-31
CRM457-1549: Add "Back calculation" autogrant limits

jsugarman:
## Description of change
Add "Back calculation" autogrant limits

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1549)
[Related bug fix ticket](https://dsdmoj.atlassian.net/browse/CRM457-1582)

It has been agreed with the BA that it can be autogranted.
Only one item (per_item unit_type only) can ever be claimed
at a rate less than or equal to the specified london/non-london
rates. all other scenarios for this service type will not be autogranted.

~~This is a less than ideal solution for implementing
autogrant for this service type.~~

~~It's implementation is pending discussion because
the service type is an outlier in a number of respects~~

~~- It is is implemented in provider as a variable hourly or item
  base service type, requiring 2 autogrant limit records per service 
  and start date (which is unique to this service)~~

~~- Only one (hour or item) can ever be claimed at a specific rate but the
  autogrant limit mechanism currently works on a "less than or equal"
  basis.~~

## Notes for reviewer
There should be 118 autogrant limits on this branch if seeding succeeded. spoiler it did.

This switches from using migrations for data creation/updating to using seeds. This is more idiomatic IMO and handles this [bug](https://dsdmoj.atlassian.net/browse/CRM457-1582) that occurs due to our branch builder setup

### DIFF
--- a/db/migrate/20240308140314_create_autogrant_limits.csv
+++ b/db/migrate/20240308140314_create_autogrant_limits.csv
@@ -2,6 +2,7 @@ service,start_date,unit_type,max_units,max_rate_london,max_rate_non_london,trave
 A&E consultant,2012/12/2,per_hour,8,108.00,100.80,40.00,40.00,4
 Accident reconstruction,2012/12/2,per_hour,15,54.40,72.00,27.20,36.00,4
 Animal behaviourist,2012/12/2,per_hour,13,85,85,40,40,4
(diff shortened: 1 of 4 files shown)


dwhenry (approved):
so long as the seeds are loaded in production...

PR #503, opened 2024-05-31
[CRM457-1543] Allow part grant without comment

dwhenry:
## Description of change
Allow part grant without a comment

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1543)


.env.development
```diff
@@ -40,11 +40,11 @@ ENABLE_SYNC_TRIGGER_ENDPOINT=true
...
-CASEWORKER_CLIENT_ID=not-a-real-key
-CASEWORKER_CLIENT_SECRET=not-a-real-key
+APP_STORE_CLIENT_ID=
+APP_STORE_TENANT_ID=
+CASEWORKER_CLIENT_ID=
+CASEWORKER_CLIENT_SECRET=
 
 # ingress and protocol for webhook subscription to app store - must match local apps's rails server
-HOSTS=localhost:3002
+HOSTS=https://localhost:3001
```
jsugarman, line 49: This is used with PROTOCOL below to generate URL (see app/services/app_store_subscriber.rb:9-10). and the port was set to 3002 because submit is set to 3001 so you can have then all running on different ports.- certainly this setup works for me.

dwhenry: updated

jsugarman (approved):

PR #509, opened 2024-06-03
CRM4 Caseworker: Calculation does not update when no. of items are adjusted

ivanELEC:
## Description of change

 [CRM4 Caseworker: Calculation does not update when no. of items are adjusted](https://dsdmoj.atlassian.net/browse/CRM457-1570)

- Account for formatted currency strings when calculating costs

Note: The changes to the quote model aren't necessarily related to the ticket, I just felt it was better to be more explicit with the cost type code, if the cost type isn't entered something should rightfully go wrong, we shouldn't assume it's per hour


ivanELEC:
Putting this back into draft as the solution for the bug on quotes doesn't fundementally fix the problem and leaves us open to setting values to 0 when they are null and actually should cause issues

jsugarman (approved):
👍 

I have compared how provider does the calculation and it is very different as it uses multi step form and actually submits/commits the data and refreshes the page. This is more "progressive" as it does not rely on javascript and centralises the calculation to just be using the ruby code, which can handle the commas in the number. But if we were to implement in assess it opens up avenues of creating events every time the user clicks "Calculate my changes". So unless we want to go down the rabbit hole that that involves handling all that logic the javascript only mechanism would seem to be the solution, as Ivan has done.

PR #519, opened 2024-06-05
Ensure state is also set in JSON blob

dwhenry:
## Description of change

 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1586)


patrick-laa (approved):
It kinda creeps me out having attributes in the blob that are also outside the blob. We've already been tripped up that there's a `created_at` field in the blob which we were accidentally reading when we expected to read the caseworker `created_at` field. BUT I do appreciate that if it's in there, we want it to be correct, and it's helpful to have it at our fingertips for the historical record.

PR #520, opened 2024-06-05
CRM457-1549: ensure seeds are run on hosted envs

jsugarman:
## Description of change
Run seeds on deploy 

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1549)

For DEV branches db:prepare will run seeds when it first creates the database, but not thereafter. hosted envs will never run db:seeds as part of db:prepare so we add it explcitly. For local development we add amigration just to ensure developers have latest seeds locally. Seeds are idempotent so this should not matter.

- Run seeds on app start for hosted environments
- Add migration to [re]seed autogrant limits locally

## Notes for reviewer
Instead of the migration we could instead rely on developers to run db:seeds themselves locally but i quite like have the migration for historical reasons.

### DIFF
--- a/db/migrate/20240605125438_seed_autogrant_limits.rb
+++ b/db/migrate/20240605125438_seed_autogrant_limits.rb
@@ -1,0 +1,5 @@
+class SeedAutograntLimits < ActiveRecord::Migration[7.1]
+  def up
+    load Rails.root.join("db/seeds/autogrant_limits.rb")
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_03_13_123836) do
+ActiveRecord::Schema[7.1].define(version: 2024_06_05_125438) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "plpgsql"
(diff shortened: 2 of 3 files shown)


db/migrate/20240605125438_seed_autogrant_limits.rb
```diff
@@ -0,0 +1,5 @@
+class SeedAutograntLimits < ActiveRecord::Migration[7.1]
+  def up
+    load Rails.root.join("db/seeds/autogrant_limits.rb")
+  end
```
dwhenry, line 4: if we need to do this locally to keep it insync do we actually need the overhead of running db:seed on startup???

I think we should choose one or the other approach. :) 

i.e. dev need to manually seed locally or reseed is done via a DB migration as we don;t need to seed on startup. :)

dwhenry (approved):
Not blocking this but 100% on board either. :)

db/migrate/20240605125438_seed_autogrant_limits.rb
jsugarman: yes its all the different scenario that bother me and don't know a way to achieve a good balance
- a new dev or a Dev branch runs db:prepare and this will seed anyway
- an existing dev will not run db:prepare or run.sh so relies on migrations to do it transparently or db:seed to do it locally
- a deployment would now work because of the db:seed in the run.sh but will also unnecessarily run the migration

It would be nice if to change the seeds we just add/amend the seeds and everything else just works.

Will post a question in our tech channel

PR #525, opened 2024-06-11
CRM4 - Content/Calculation Change - Service Type - Translation (documents)

ivanELEC:
## Description of change

 [CRM4 - Content/Calculation Change - Service Type - Translation (documents)](https://dsdmoj.atlassian.net/browse/CRM457-1612)

- Update quote view model to consume extra fields from provider app (cost_multiplier, cost_item_type)
- Update service costs with adjustments model view to consume and use cost multiplier
- Update views that include service cost (cost fields, service costs)
- Update CostAdjustment javascript to multiply item totals by cost multiplier (unless additional cost)
- Update tests

### DIFF
--- a/app/javascript/component/prior_authority/cost_adjustment.js
+++ b/app/javascript/component/prior_authority/cost_adjustment.js
@@ -65,12 +65,12 @@ export default class CostAdjustment {
       let costPerItemNum = convertCurrencyToNumber(this.costPerItemField?.value)
       let items = parseInt(this.itemsField?.value)
       let unitPrice = new Decimal(costPerItemNum)
-      let multiplier = new Decimal(this.itemsField?.getAttribute('data-cost-multiplier'))
       let unrounded = null;
       if(this.additionalCost){
         unrounded = unitPrice.times(items);
       }
       else{
+        let multiplier = new Decimal(this.itemsField?.getAttribute('data-cost-multiplier'))
         unrounded = unitPrice.times(items).times(multiplier);
       }
       return unrounded.toFixed(2);
--- a/app/javascript/component/prior_authority/cost_adjustment.js
+++ b/app/javascript/component/prior_authority/cost_adjustment.js
@@ -68,10 +68,10 @@ export default class CostAdjustment {
       let multiplier = new Decimal(this.itemsField?.getAttribute('data-cost-multiplier'))
       let unrounded = null;
       if(this.additionalCost){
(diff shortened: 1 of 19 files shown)


app/javascript/component/prior_authority/cost_adjustment.js
```diff
@@ -64,8 +65,14 @@ export default class CostAdjustment {
...
       let items = parseInt(this.itemsField?.value)
       let unitPrice = new Decimal(costPerItemNum)
-
-      let unrounded = unitPrice.times(items);
+      let unrounded = null;
+      if(this.additionalCost){
+        unrounded = unitPrice.times(items);
+      }
+      else{
+        let multiplier = new Decimal(this.itemsField?.getAttribute('data-cost-multiplier'))
```
jsugarman, line 73: is the question mark some kind of safe navigator operator in JS? `itemsField?`

app/view_models/prior_authority/v1/quote.rb
```diff
@@ -9,6 +9,8 @@ class Quote < BaseWithAdjustments
       attribute :id, :string
       attribute :cost_type, :string
       attribute :item_type, :string, default: 'item'
+      attribute :cost_item_type, :string
+      attribute :cost_multiplier, :decimal, precision: 10, scale: 5, default: 1
```
jsugarman, line 13: why does the scale need to be greater than for cost per hour/item (scale: 2)?

jsugarman: scratch that i assume that is because 0.001, although is that  not `scale: 3`. but sounds ok anyway

spec/view_models/prior_authority/v1/quote_spec.rb
```diff
@@ -4,14 +4,22 @@
...
+    context 'when cost is per item and default cost multiplier' do
       let(:attributes) { { cost_type: 'per_item', items: 3, cost_per_item: '33.5' } }
 
       it 'returns total item cost' do
         expect(subject).to eq 100.5
       end
     end
 
+    context 'when cost is per item and custom cost multiplier' do
+      let(:attributes) { { cost_type: 'per_item', items: 3, cost_per_item: '33.5', cost_multiplier: 0.1 } }
```
jsugarman, line 16: should we not test with the "kiloword" multiplier, `0.001`?

1/0.001 = 1000.0

jsugarman (approved):
👍 looks ok to me but have not tried it out.

PR #534, opened 2024-06-14
CRM457-1478: Embedded metabase charts

patrick-laa:
## Description of change
- Add a new route that only supervisors can access or see link to
- That route uses the embedding secret key to generate a JWT
- Then it creates a URL with the right metabase host, dashboard ID and the JWT and uses that to populate an iframe
 [Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1478)

Screenshot 2024-06-14 at 12 49 08


app/controllers/dashboards_controller.rb
```diff
@@ -0,0 +1,18 @@
+class DashboardsController < ApplicationController
+  before_action :authorize_supervisor
+
+  def show
+    payload = {
+      resource: { dashboard: ENV.fetch('METABASE_DASHBOARD_ID').to_i },
+      params: {},
+      exp: Time.now.to_i + (60 * 10) # 10 minute expiration
```
jogilder, line 8: why 10 mins expiry?

helm_deploy/templates/deployment.yaml
```diff
@@ -191,6 +191,16 @@ spec:
...
             - name: ENABLE_PROMETHEUS_EXPORTER
               value: 'true'
+            - name: METABASE_DASHBOARD_ID
+              value: '{{ .Values.metabase.dashboardId}}'
+            - name: METABASE_SECRET_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: app-secrets
+                  key: metabase_secret_key
+            - name: METABASE_SITE_URL
```
jogilder, line 201: Wondering if we should just not add the metabase siteurl into the metabase secret? Maybe this gives us more flexibility for testing?

helm_deploy/templates/deployment.yaml
```diff
@@ -191,6 +191,16 @@ spec:
               value: 'true'
             - name: ENABLE_PROMETHEUS_EXPORTER
               value: 'true'
+            - name: METABASE_DASHBOARD_ID
```
jogilder, line 194: It is possible that we could have more than one dashboard. Either that or we need to try several charts/tables in one dashboard. 
Do you get a separate secret for each dashboard? Or do we get one and then need a list of IDs?

jogilder (approved):
Looks good!
Couple of comments.

So the current dashboard question doesn't seem to work if you choose anything other than today. Not sure whether that is because it doesn't work on metabase/there is no data or because the parameters are not set correctly.

Couple of other comments about trying more than one or two charts but otherwise looking good.

app/controllers/dashboards_controller.rb
patrick-laa: It was the default set in the copy-pastable embed code. But I've checked and the expiration only needs to be as long as it takes the iframe to initially load, not to cover subsequent playing with widgets. So I've reduced to 10 seconds.

helm_deploy/templates/deployment.yaml
patrick-laa: I'm not sure about this one - I get the idea about flexibility, but I don't think we want to start muddying up secrets and non-secrets. Ideally there'd be some UI-configurable config set-up for non-secret env vars to handle things like this.

helm_deploy/templates/deployment.yaml
patrick-laa: I'm making it so that we use a list of IDs and render them one by one.

app/controllers/dashboards_controller.rb
patrick-laa: I take it back - if you have the page open for a long time then play with the widget date selector it stops working after some built-in grace period. So I'll set this to an hour to make sure the data doesn't stop working mid-browse.

PR #536, opened 2024-06-14
Remove coverage skip in CostItemTypeDependantValidator

ivanELEC:
## Description of change

 [Caseworker: Remove Coverage Skip for CostItemTypeDependantValidator](https://dsdmoj.atlassian.net/browse/CRM457-1620)

We do not use the pluralize feature and for cost item type validation and the way it was written for item type validation won't work as cost_item_type_for inputs into a translation - therefore, I removed that part of the code completely. If we need to use pluralize in the future, we can update appropriately

### DIFF
--- a/app/forms/prior_authority/base_cost_adjustment_form.rb
+++ b/app/forms/prior_authority/base_cost_adjustment_form.rb
@@ -11,7 +11,7 @@ module PriorAuthority
     attribute :cost_per_hour, :gbp
 
     with_options if: :per_item? do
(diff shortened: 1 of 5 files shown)


patrick-laa:
So it looks like now we've split up ItemTypeDependantValidator and CostItemTypeDependantValidator, the former is _always_ pluralized and the latter is _never_ pluralized. So if we're going to properly tidy this up, I think we should remove the `pluralize` option from ItemTypeDependantValidator as well, because it's always true, so we can go from 1 complex validators to 2 simple ones.

jsugarman (approved):

PR #545, opened 2024-06-18
Use assess_a_crime_form layout for footer links

ivanELEC:
## Description of change

[Submit a crime form provider, CRM4 provider and CRM7 provider links in footer](https://dsdmoj.atlassian.net/browse/CRM457-1631)


app/controllers/about/accessibility_controller.rb
```diff
@@ -2,6 +2,8 @@ module About
   class AccessibilityController < ApplicationController
     skip_before_action :authenticate_user!
 
+    layout 'assess_a_crime_form'
```
patrick-laa, line 5: If we're fixing this I sort of feel like we should fix it properly. At the moment we have 4 layouts:
`application` - specific to NSM, applied by default
`assess_a_crime_form` - generic, but only used if explicitly specified
`prior_authority` - specific to Prior Authority, only used if explicitly specified
`external` - minimal and specific to NSM, only used if explicitly specified.

That doesn't feel right to me. I think we should _default_ to `assess_a_crime_form`, and only switch to NSM-an specific layout if we're looking at NSM-specific content. So if it was me I'd:

rename `assess_a_crime_form.html.erb' to 'application.html.erb' so it gets used by default
rename 'application.html.erb' to 'nsm.html.erb', create a new controllers/nsm/base_controller.rb, have all nsm controllers inherit from that, and in that specify that we're using the nsm layout
delete external.html.erb as I think assess_a_crime_form.html.erb covers it just fine.

ivanELEC: good suggestion - made these changes

patrick-laa (approved):

PR #551, opened 2024-06-19
Add separate replica counts for app/web and worker

jsugarman:
## Description of change
Add separate replica counts for app/web and worker

we do not need as many workers pods as app pods typically
but these should be seprate in any event.


ivanELEC (approved):

patrick-laa (approved):
Love it, but can you hold off on merging for 10 minutes or so while I hand-hold the deployment of the previous PR?

PR #543, opened 2024-06-18
Use flashier dashboard on dev

patrick-laa:
Embed Mary's production dashboard in dev environment
Screenshot 2024-06-18 at 11 05 48

PR #553, opened 2024-06-19
Crm457 1634 basic layout

ivanELEC:
## Description of change

[Provider: Basic Layout for Analytics Page](https://dsdmoj.atlassian.net/browse/CRM457-1634)

This is mostly placeholder stuff behind feature flags

I frame id list has been split into prior auth and nsm and all render on the 'overview' page of the respective service
navigation is all controlled and rendered via params and instance variables but this might not be the best way to do things - considering creating a presenter class for this

image
image


config/feature_flags.yml
```diff
@@ -10,10 +10,15 @@ feature_flags:
     uat: false
     production: false
   insights:
-    local: <%= ENV.fetch("INSIGHTS", false) %>
+    local: <%= ENV.fetch("INSIGHTS_ENABLED", false) %>
     development: true
     uat: false
     production: false
+  nsm:
```
patrick-laa, line 17: I don't like the name of this feature flag - NSM absolutely is enabled, and caseworkers can see the NSM area of the app when they log in. Can it be `nsm_insights`?

app/views/layouts/dashboard.html.erb
```diff
@@ -0,0 +1,43 @@
...
+
+<% content_for(:content) do %>
+  <%= yield %>
+<% end %>
+
+<% content_for(:footer_links) do %>
+  <% render partial: 'layouts/footer_links' %>
+<% end %>
+
+<%# render partial: 'layouts/developer_tools' if FeatureFlags.developer_tools.enabled? %>
```
patrick-laa, line 41: I don't think we have a partial called this - I think this is commented out in every layout, and we should probably ditch it.

helm_deploy/values-dev.yaml
```diff
@@ -119,7 +119,8 @@ redis:
     password: redis
 
 metabase:
-  dashboardIds: 14
+  priorAuthorityDashboardIds: 14
```
patrick-laa, line 122: Mary asked if we can use her slightly snazzier dashboard in dev so we can see it. I created a PR, but maybe we can sneak it in here? https://github.com/ministryofjustice/laa-assess-crime-forms/pull/543

patrick-laa (approved):

PR #558, opened 2024-06-20
CRM457-1650: Standardise modsec and add correct tagging

patrick-laa:
## Description of change
Instead of 3 copies of the same config, just use 1
Add the annotations that allow us to view modsec logs

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1650)


jsugarman (approved):
tried it out with a dry run and it looks good to me. can't speak to the modesec rules though

PR #572, opened 2024-06-25
Crm457 1665/send eveents

dwhenry:
## Description of change
Push reportring events (Assignment, Unassignment and NewVersion to app store as they are created. Other events are synced once a decision/RFI request is made.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1665)

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -55,6 +55,10 @@ class Event < ApplicationRecord
 
     private
 
+    def notify(event)
+      NotifyAppStore.perform_later(event)
+    end
+
     def create_dummy_users_if_non_production(params)
       create_dummy_user_if_non_production(params['primary_user_id'])
       create_dummy_user_if_non_production(params['secondary_user_id'])
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -55,10 +55,6 @@ class Event < ApplicationRecord
 
     private
 
-    def notify(event)
-      NotifyEventAppStore.perform_later(event:)
-    end
-
     def create_dummy_users_if_non_production(params)
       create_dummy_user_if_non_production(params['primary_user_id'])
       create_dummy_user_if_non_production(params['secondary_user_id'])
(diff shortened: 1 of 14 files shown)


app/models/event.rb
```diff
@@ -55,6 +55,10 @@ def latest_provider_update
 
     private
 
+    def notify(event)
```
patrick-laa, line 58: I don't think this needs an argument, does it? You could just call `NotifyEventAppStore.perform_later(event: self)`, make the method signature much more straightforward.

app/services/app_store_client.rb
```diff
@@ -40,6 +40,17 @@ def create_subscription(payload)
...
   end
 
+  def create_event(submission_id, payload)
+    response = self.class.post("#{host}/v1/submissions/#{submission_id}/events", **options(payload))
+
+    case response.code
+    when 200..204
+      :success
+    else
+      raise "Unexpected response from AppStore - status #{response.code} on create subscription"
```
patrick-laa, line 50: `subscription` should be `event`

app/services/app_store_client.rb
```diff
@@ -40,6 +40,17 @@ def create_subscription(payload)
     end
   end
 
+  def create_event(submission_id, payload)
```
patrick-laa, line 43: Nit: If the AppStoreClient has a `create_event` method, that suggests that method should only be usable to create an event. But you could in theory call `create_event` with an arbitrary length array of events. I reckon either this should be called `create_event`, or it should take a _single_ event object as its second argument and create a payload from that.

patrick-laa (approved):

app/models/event.rb
dwhenry: class method.. :)

app/services/app_store_client.rb
dwhenry: I think I like change the name to `create_events` to avoid this method needing to know about payload creation...

app/models/event.rb
dwhenry: converted to an instance method

PR #600, opened 2024-07-08
CRM457-1557: Add search UI

patrick-laa:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1557)

## Screenshots of changes:
Screenshot 2024-07-08 at 14 54 52

... Wait for a back-end or hack the AppStoreClient to return a payload similar to the specs.

### DIFF
--- a/app/controllers/prior_authority/searches_controller.rb
+++ b/app/controllers/prior_authority/searches_controller.rb
@@ -1,19 +1,19 @@
 module PriorAuthority
   class SearchesController < BaseController
     def show
-      @search = Search.new(search_params)
-      @search.execute if @search.valid?
+      @search_form = SearchForm.new(search_params)
+      @search_form.execute if @search_form.valid?
     end
 
     def new
-      @search = Search.new(default_params)
+      @search_form = SearchForm.new(default_params)
       render :show
     end
 
     private
 
     def search_params
-      params.require(:search).permit(
+      params.require(:search_form).permit(
         :query,
         :submitted_from,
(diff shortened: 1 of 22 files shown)


ivanELEC (approved):

app/controllers/prior_authority/searches_controller.rb
```diff
@@ -0,0 +1,36 @@
+module PriorAuthority
+  class SearchesController < BaseController
+    def show
+      @search = Search.new(search_params)
```
dwhenry, line 4: so every other one of these is called `SearchForm`.. whould we do the same here?

app/models/search_result.rb
```diff
@@ -0,0 +1,39 @@
+class SearchResult
```
dwhenry, line 1: should this live under `view_models` as it is a model built from the JSON response?

app/models/search_result.rb
```diff
@@ -0,0 +1,39 @@
...
+  def submission
+    @submission ||= Submission.find_by(id: @data[:application_id]) || UpdateSubmission.call(@data.deep_stringify_keys)
+  end
+
+  def laa_reference
+    submission.data['laa_reference']
+  end
+
+  def firm_name
+    submission.data['firm_name']
```
dwhenry, line 20: any reason we can't look this up form the response JSON? try and avoid touching the local DB as much as possible?

dwhenry (approved):
I like it

PR #601, opened 2024-07-08
CRM457-1726: Swallow event sync errors if they are redundant

patrick-laa:
## Description of change
Sometimes individual event syncs fail because of a race condition that's harmless because the event has already been bulk synced to the app store. If so, log a warning and allow the job to pass instead of raising an error

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1726)


dwhenry (commented):
Would it not make more sense to add this code to the app store.. so that when receiving event data - it returns a valid state even if the application is locked if the event already exists?

ivanELEC (approved):

PR #603, opened 2024-07-09
CRM457-1660: Use internal hostname to bypass modsec

patrick-laa:
## Description of change
By using the service URL instead of the host name associated with the ingress to construct our webhook url, we allow our app store -> caseworker HTTP requests to bypass modsec, which is good because modsec is notorious for false positives, and these requests should never be treated as a modsec positive.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1660)

## Notes for reviewer
I have adjusted our Cloud Platform permissions to allow internal network requests from the app store namespaces to the caseworker namespaces.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -47,4 +47,3 @@ CASEWORKER_CLIENT_SECRET=not-a-real-key
 
 # ingress and protocol for webhook subscription to app store - must match local apps's rails server
 HOSTS=localhost:3002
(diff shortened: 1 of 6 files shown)


jsugarman (approved):
makes sense since we are already calling the app store via its internal service url for pull_updates and usage of AppStoreClient  from what i can see.

PR #619, opened 2024-07-15
Crm457 1723/updated crm7 sumamry tables

dwhenry:
## Description of change
update summary tables
[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1723)


jsugarman (dismissed):

config/locales/en/nsm/work_items.yml
```diff
@@ -30,6 +30,11 @@ en:
         allowed_cost: Net cost allowed
         total: Total
         table_info_item: work item
+        summary_table_caption: Summary of claimed work items
+        table_caption: Claimed work items
+        accessible:
+          claimed: 'Sum of net cost claimed: '
```
jsugarman, line 36: since this is read by screen readers should it say "Sum of net cost claimed/allowed is"?

jsugarman (approved):

config/locales/en/nsm/work_items.yml
ivanELEC: worth noting not all screen readers will read the slash so "claimed or allowed" might be better? But maybe we can have it like this for now, ask content and change later?

ivanELEC (approved):

config/locales/en/nsm/work_items.yml
dwhenry: so in this instance it is only claimed... but this is more a question for chris/mary I would expect...

PR #624, opened 2024-07-17
Enable analytics uat

ivanELEC:
To allow us to present analytics to stakeholders

### DIFF
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -83,7 +83,6 @@ redis:
 metabase:
   priorAuthorityDashboardIds: ''
   nsmDashboardIds: ''
-  priorAuthorityAutograntDashboardIds: ''
   siteUrl: 'https://crime-forms-metabase.service.justice.gov.uk'
 
 customErrorService:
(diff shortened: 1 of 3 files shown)


helm_deploy/values-prod.yaml
```diff
@@ -83,7 +83,6 @@ redis:
 metabase:
   priorAuthorityDashboardIds: ''
   nsmDashboardIds: ''
-  priorAuthorityAutograntDashboardIds: ''
```
patrick-laa, line 86: Are you sure it's safe to remove this? `deployment.yaml` still references it when setting `METABASE_PA_AUTOGRANT_DASHBOARD_IDS` (which is admittedly unused). I think helm might break if a referenced key doesn't exist in a values file.

ivanELEC: good shout - will remove from helm too

patrick-laa: Cool, then for completeness it would be worth removing `METABASE_PA_AUTOGRANT_DASHBOARD_IDS` from dashboard_controller_spec, and `priorAuthorityAutograntDashboardIds` from values-dev.yaml

ivanELEC: updated

patrick-laa (approved):

PR #627, opened 2024-07-17
Return correct uplift when value is nil

elken:
When editing a work item, if the uplift is nil it doesn't display anything in the provider submission table.

Using type coercion, we can convert the false-y nil value to 0.

### Before changes:
image

### After changes:
image


jsugarman (approved):

robinsk1 (approved):

elken:
I'll rebase and merge this tomorrow

PR #628, opened 2024-07-18
CRM457-1757: Display last_updated date on search results

jsugarman:
## Description of change
*TBD*

Display last_updated date on search results

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1757)

This is one option for handling edge cases where a submissions
updated_at goes out sync with events.

The "last_updated" date is the actual date that is both filtered and sorted
on by the appstore searches API and should therefore match the search results
returned exactly.

This option does, however, create a divide between the search results page
and other index pages that show "your", "open" and "closed" applications
sorted|sortable by submission.updated_at.

## Alternative
Another option is to sync the top-level`last_updated_at` on appstore wtih submissions
on caseworker. However, aside from adding the column to submissions in caseworker, backfilling 
it and syncing via the UpdateSubmission service we would also need to handle the fact that an event
would need to update it in caseworker, as it does in appstore.

### DIFF
--- a/app/view_models/search_result.rb
+++ b/app/view_models/search_result.rb
@@ -2,8 +2,9 @@ class SearchResult
   include NameConstructable
   include SubmissionTagHelper
 
-  def initialize(data)
-    @data = data.deep_stringify_keys
+  def initialize(view_data, raw_data)
+    @view_data = view_data.deep_stringify_keys
+    @data = raw_data.deep_stringify_keys
   end
 
   delegate :id, to: :submission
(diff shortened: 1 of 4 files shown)


app/view_models/search_result.rb
```diff
@@ -29,8 +30,8 @@ def caseworker
     submission.assignments.first&.display_name || I18n.t('search.unassigned')
   end
 
-  def date_updated
-    submission.updated_at.to_fs(:stamp)
+  def last_updated
```
patrick-laa, line 33: I would much prefer a solution that solves this for the 4 places that use submission.updated_at all together, so that we don't risk having the caseworker app show different dates for the same application in different places. Having the search result show one thing, and the application details page (and the open/closed/your applications table) show another, is not ideal.

That being said, now that we have a viable search endpoint, maybe the ideal thing would be for the other tables to actually be populated by the search endpoint, and frankly the application details view to be populated by it too...

ivanELEC:
Closing due to opting for alternate solution: 
https://github.com/ministryofjustice/laa-assess-crime-forms/pull/630

PR #630, opened 2024-07-19
CRM457-1757: Make sure "last updated" is accurate

patrick-laa:
## Description of change
Alternative approach to https://github.com/ministryofjustice/laa-assess-crime-forms/pull/628. This will make 'last updated' almost entirely consistent with the app store value used to filter search results, while keeping it entirely consistent across the caseworker app.

I think the 'almost' above is such an extremely edgy edge case that it's ok for now, and the long-term solution is to pull data directly from the app store instead of a local database for all list and details views of applications.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1757)

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -24,7 +24,9 @@ class Event < ApplicationRecord
   LOCAL_EVENTS = [
     'Event::ChangeRisk',
     'Event::DraftDecision',
+    'Event::Edit',
     'Event::Note',
+    'Event::UndoEdit',
     'PriorAuthority::Event::DraftSendBack',
   ].freeze
 
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -20,7 +20,16 @@ class Event < ApplicationRecord
     'PriorAuthority::Event::DraftSendBack',
     'PriorAuthority::Event::SendBack',
   ].freeze
+
+  NOTIFYING_EVENTS = [
+    'Event::Assignment',
+    'Event::Unassignment',
+    'Event::NewVersion',
+  ].freeze
+
   scope :history, -> { where(event_type: HISTORY_EVENTS).order(created_at: :desc) }
+
+  scope :notifying, -> { where(event_type: NOTIFYING_EVENTS) }
 
   # simplifies the rehydrate process
   attribute :public
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -21,15 +21,16 @@ class Event < ApplicationRecord
     'PriorAuthority::Event::SendBack',
   ].freeze
 
-  NOTIFYING_EVENTS = [
-    'Event::Assignment',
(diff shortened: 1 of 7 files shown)


jsugarman (commented):
👍 definitely neater although having the logic for "last updated" in two places does concern me long term. 

Why have you limited last_updated it to certain events? The last_updated_at in appstore is typically either, for a brief window, the date a submission is created, otherwise the latest event received by the appstore's `created_at` regardless. [here](https://github.com/ministryofjustice/laa-crime-application-store/blob/main/app/services/submissions/update_service.rb#L26-L28)

patrick-laa:
@jsugarman re this bit:

> Why have you limited last_updated to certain events? 

It's because a bunch of other events created locally will _not_ affect the app store last_updated_at, e.g if I add a note there will be local Event::Note whose created_at will be _later_ than the app store last_updated_at. That event will only make it to the app store when the app is updated with a new version.

Ah. But at that point there won't be an event created locally because a caseworker updating the app store with a send-back of assessment, even though it creates a new version, won't create a NewVersion event, will it?

So really I need to extend NOTIFYING_EVENTS to include all events that immediately get synced to the app store, i.e. everything except DraftDecision, Note and DraftSendBack.

patrick-laa:
Ok, now updated. Re this bit:

> having the logic for "last updated" in two places does concern me

Me too. Ultimately we want only the app store to calculate it. We could make sure we sync it back to the caseworker app every time it changes, but to my mind the best approach is to load in this value directly from the app store whenever we want to display it, i.e. add it to the standard payload and get that from the app store on each page load.

jsugarman:
@patrick-laa 
> 
> Ah. But at that point there won't be an event created locally because a caseworker updating the app store with a send-back of assessment, even though it creates a new version, won't create a NewVersion event, will it?
> 
> So really I need to extend NOTIFYING_EVENTS to include all events that immediately get synced to the app store, i.e. everything except DraftDecision, Note and DraftSendBack.

yeah, I was looking at events data in appstore and trying to identify which and where assess was notifying appstore. "decision" and "sent_back" events, at least, stuck out as ones that I would expect to update "last updated" (in assess as in appstore).

jsugarman (approved):
Approving as a less than ideal solution but one which that provides consistency across caseworker index-like views.

@ivanELEC It may be worth creating a new ticket off the back of the comment below,  or looking into it as an alternative to this PR's approach. Am approving regardless.

>Ok, now updated. Re this bit:

having the logic for "last updated" in two places does concern me

> ... Ultimately we want only the app store to calculate it. We could make sure we sync it back to the caseworker app every time it changes, but to my mind the best approach is to load in this value directly from the app store whenever we want to display it, i.e. add it to the standard payload and get that from the app store on each page load.

PR #632, opened 2024-07-19
Crm457 1577 analytics search 2

ivanELEC:
## Description of change

[Analytics Search](https://dsdmoj.atlassian.net/browse/CRM457-1639)


app/forms/search_form.rb
```diff
@@ -4,6 +4,11 @@ class SearchForm
   include ActiveRecord::AttributeAssignment
 
   Option = Struct.new(:value, :label)
+  ApplicationType = Struct.new(:id, :name)
```
patrick-laa, line 7: I feel like since all we need is a thing with a value and a label, we could reuse the `Option` struct that does exactly that rather than make another struct that does exactly the same thing with different words.

app/forms/search_form.rb
```diff
@@ -4,6 +4,11 @@ class SearchForm
...
+  APPLICATION_TYPES = [
+    ApplicationType.new('crm4', 'Prior authority'),
```
patrick-laa, line 9: These user-facing strings should be in locale files, not hard-coded into the model

config/locales/en/dashboards.yml
```diff
@@ -5,4 +5,10 @@ en:
...
         prior_authority: Prior authority
         nsm: Non-standard magistrates
-        search: Search for an application
+        search: Search
+    search:
+      application_type:
+        label: Which service do you want to search?
+      query:
+        label: Claim or application details
+        hint: Search for a client/ defendant name, firm name, UFN or LAA reference
```
patrick-laa, line 14: I suspect there shouldn't be a space after the slash

app/forms/search_form.rb
```diff
@@ -58,21 +64,28 @@ def statuses
...
-    field_set = [:query, :submitted_from,
-                 :submitted_to, :updated_from,
-                 :updated_to, :status_with_assignment,
-                 :caseworker_id].any? do |field|
+    fields = [:query, :submitted_from,
+              :submitted_to, :updated_from,
+              :updated_to, :status_with_assignment,
+              :caseworker_id]
+
+    fields.append(:application_type) if form_context == 'analytics' && FeatureFlags.nsm_insights.enabled?
```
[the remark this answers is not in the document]
ivanELEC: i agreed with you and then realised that this depends on what logic is desired for the page
application_type is always present in a service context, and it's one of the fields that at_least_one_field_set checks for, then the form will be considered valid if you don't enter details - this is fine if we think it's ok that not entering anything should return all crm4 applications - but I had made an assumption that users would need to apply a filter

In the same vein, if no filter in service context returns all crm4s, should no filter in analytics context return all submissions?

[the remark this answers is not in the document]
patrick-laa: Ohh, yeah my bad, I misread my own code! Yes, we do need users to enter something on the Prior Authority search page, there is an explicit requirement that if they leave everything blank we show them an error message. Whereas in the analytics context, if they leave everything blank _except_ the application type, then it might be we don't want to show them an error. So depending on context, application type potentially should be one of the "at least one of these fields must be set".

If that's the route we go down, then I don't think the SearchForm should _know_ about what the different contexts are - it shouldn't know that the requirements for the analytics screen are different to the PA search screen. Instead, I think in those contexts should explicitly tell the SearchForm what their requirements are. E.g. instead of the controller telling SearchForm "this is the analytics context, I expect you to know the ramifications of that", it should tell it "this is a context where application_type is not one of the fields that at least one of needs to be set".

So maybe it's an "explicit_application_type" boolean rather than a "form_context" string?

app/views/shared/_search.html.erb
```diff
@@ -0,0 +1,128 @@
...
+            <div class="govuk-form-group">
+              <%= f.govuk_date_field :updated_to, legend: { size: 's', text: t('.updated_to') } %>
+            </div>
+          </div>
+        </div>
+        <div>
+          <div class="govuk-date-input__item">
+            <div class="govuk-form-group">
+              <%= f.govuk_collection_select :status_with_assignment,
+                                            @search_form.statuses,
```
patrick-laa, line 50: Slight issue here - in prior authority, sent back applications have status `sent_back`, in NSM they have status `further_info`. As it stands, this dropdown only has `sent_back`. It _might_ be we should update it so that we use `sent_back` for NSM too, otherwise as it stands we don't have a single status that we can search on to find both NSM _and_ PA sent back submissions.

ivanELEC: makes sense - if anything, further_info could apply to both but doesn't necessarily whereas sent_back works for both. For now I'll separate that out into a different ticket in case we want to discuss it as a team properly and decide on what's most appropriate

PR #637, opened 2024-07-22
Ensure that turbo frames render during tests

elken:
## Description of change

Noticed during #635, turbo frames don't render inside the Capybara environment without these helpers enabled.

[Discussion](https://discuss.hotwired.dev/t/turbo-frame-not-works-with-capybara/3990)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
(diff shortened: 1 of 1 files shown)


dwhenry:
I would approve but we are about to remove all the turboframe related stuff... so do we still need this?

elken:
> I would approve but we are about to remove all the turboframe related stuff... so do we still need this?

Oh is that the only place we use it? Probably not then ha

PR #646, opened 2024-07-25
CRM457-1790: Discard  all adjustments prior to sending back

patrick-laa:
## Description of change
If I make adjustments then send an application back, those adjustments are effectively ignored, in that the provider app does not pick them up and when the provider app pushes a new version to the app store, those adjustments are overwritten anyway. BUT there's an unintended side effect, which is that when trying to work out which fields the provider has changed, the provider app diffs its database with the latest app store version. It picks up on the adjustments in the diff, with the end result that in the caseworker UI the fields the caseworker previously adjusted end up being highlighted as ones the _provider_ has adjusted in the RFI loop.

This PR undoes caseworker adjustments when an application is sent back, so that the version available in the app store has no adjustments, so the "what has the provider changed?" diff no longer gets messed with.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1790)

### DIFF
--- a/app/forms/prior_authority/send_back_form.rb
+++ b/app/forms/prior_authority/send_back_form.rb
@@ -30,6 +30,7 @@ module PriorAuthority
       return false unless valid?
 
       PriorAuthorityApplication.transaction do
+        discard_all_adjustments
         stash(add_draft_send_back_event: false)
         update_local_records
         NotifyAppStore.perform_later(submission:)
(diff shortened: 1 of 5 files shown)


dwhenry (approved):
Works, but lots of moving parts...

did you consider just reloading the data from the app store to reset the main blob... this has the advantage that this code is much simpler and you don;t have to update it when the data structure changes.. :) 

as for the undo events... could this just be created from..

```
events_to_undo = submission.events.where(event_type: 'Event::Edit', version: submission.version)
events_to_undo....
```

PR #647, opened 2024-07-25
[CRM457-1753] Ensure table rows act like anchors

elken:
## Description of change

This enables a workflow of clicking on a work item or disbursement from a table and being able to navigate back directly to the row in question.

I've had to adjust the back buttons thanks to [how `link_to :back`](https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to) interacts with the Content-Security-Policy, but if there's a nicer way to handle this then I'd love to know how.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1753)

### DIFF
--- a/app/javascript/lib/table-item-anchors.js
+++ b/app/javascript/lib/table-item-anchors.js
@@ -1,0 +1,12 @@
+document.addEventListener("DOMContentLoaded", function () {
+  [...document.querySelectorAll("table.govuk-table tr[id] a")].map(
+    function (link) {
+      link.addEventListener("click", function () {
+        const nearestRow = link.closest("tr");
+        if (nearestRow) {
+          history.replaceState({}, "", `#${nearestRow.id}`);
+        }
+      });
+    },
+  );
+});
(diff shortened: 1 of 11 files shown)


app/javascript/lib/table-item-anchors.js
```diff
@@ -0,0 +1,20 @@
+document.addEventListener("DOMContentLoaded", function () {
+  [...document.querySelectorAll("a#back-button")].map(function (btn) {
```
dwhenry, line 2: any reason for using `map` instead of `forEach`??

app/javascript/lib/table-item-anchors.js
```diff
@@ -0,0 +1,20 @@
...
+    btn.addEventListener("click", function (evt) {
+      evt.preventDefault();
+      history.back();
+    });
+    return btn;
+  });
+  [...document.querySelectorAll("table.govuk-table tr[id] a")].map(
```
dwhenry, line 9: would prefer to be explicit about this I think... so
```
"table.store-anchor-on-click tr[id] a"
```

dwhenry (approved):

app/javascript/lib/table-item-anchors.js
elken: Sure I can add more specificity

dwhenry (approved):

PR #654, opened 2024-07-30
Crm457 1770/updated at

dwhenry:
## Description of change
Set the updated_at in the json blob as the data is sent back to the app store. This is useful for reporting.

Have a separate PR to update existing records in app store

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1770)


ivanELEC (approved):

ivanELEC:
Similarly to other PR - approved but could be enhanced with additional tests

PR #657, opened 2024-07-31
Rely on capybara default wait time rather than roll our own

jsugarman:
## Description of change
Rely on capybara default wait time rather than roll our own

The reason for intermittent failure of the evaluate script previously
was because no expectations were being made of the page other than an
evaluate_script call (which does not wait implicitly). 

By adding explicit capybara expectations before the evaluate_script
call we implicity make use of capybaras default timeout/retry mechanism.


ivanELEC (approved):

elken:
There needs to be an explicit wait there though as it takes time for the anchor jump to actually happen

PR #658, opened 2024-07-31
Add linting step for helm chart

ivanELEC:
## Description of change

[Last outcome for this ticket](https://dsdmoj.atlassian.net/browse/CRM457-1547)

Note: this runs helm lint AND helm template (debug) because from my experience helm lint won't always necessarily catch issues that can break a chart (I tested helm lint on some misconfigurations and it didn't catch some of them) where helm template will AND helm template doesn't give us warning level suggestions like helm lint does - so both in tandem covers all bases for breaking changes and suggestions

Also including yamllint to check for things like duplicate keys upon suggestion from @elken[ in this PR](https://github.com/ministryofjustice/laa-submit-crime-forms/pull/1013)


elken:
This doesn't check for the duplicate keys issue that was raised in the discussion though, you have to use [yamllint](https://github.com/ministryofjustice/laa-submit-crime-forms/pull/1013).

I just ran that on provider and it passes despite the duplication still existing

ivanELEC:
> This doesn't check for the duplicate keys issue that was raised in the [discussion](https://mojdt.slack.com/archives/C06HFRK36KY/p1721746330169129) though, you have to use [yamllint](https://github.com/ministryofjustice/laa-submit-crime-forms/pull/1013).
> 
> I just ran that on provider and it passes despite the duplication still existing

Indeed it doesn't, having yaml lint in addition is definitely beneficial
Is it OK if I combine your work into this PR (that we we can run yamllint against the output of each environment and not just prod)

elken:
Not a problem at all 😄

elken (approved):

PR #668, opened 2024-08-07
Crm457 1927: Fix Autogrants Task

ivanELEC:
## Description of change

[Status not being updated in JSON blob for expired/autogranted actions from CW](https://dsdmoj.atlassian.net/browse/CRM457-1927)

## Notes for reviewer
Task syncs specifically autogrant events for a comma delimited list of submission ids
The ids are going to be collected by running a companion task on app store [which is in this PR](https://github.com/ministryofjustice/laa-crime-application-store/pull/241)


lib/tasks/update_to_app_store.rake
```diff
@@ -0,0 +1,16 @@
+namespace :update_to_app_store do
+  desc 'Re-update events for a list of submissions by id'
+  task :sync_autogrant_events, [:ids] => [:environment] do |t, args|
+    id_array = args[:ids].split(',')
```
elken, line 4: `id_array` isn't used again so you can probably remove the local var

PR #672, opened 2024-08-08
CRM457-1577: Add check for application_type if it's an option

ivanELEC:
## Description of change

[CRM 4 & 7: Metabase pages](https://dsdmoj.atlassian.net/browse/CRM457-1577)

## Notes for reviewer

- This is an extra PR to address [this QA comment](https://dsdmoj.atlassian.net/browse/CRM457-1577?focusedCommentId=490733)

### DIFF
--- a/spec/system/dashboards_spec.rb
+++ b/spec/system/dashboards_spec.rb
@@ -136,8 +136,6 @@ RSpec.describe 'Dashboards', :stub_oauth_token do
                       20) + create_list(:prior_authority_application, 1, state: 'sent_back', updated_at: Date.yesterday - 1)
         end
         let(:endpoint) { 'https://appstore.example.com/v1/submissions/searches' }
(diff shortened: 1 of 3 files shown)


spec/system/dashboards_spec.rb
```diff
@@ -136,27 +136,8 @@
...
-            application_type: 'crm4',
-            explicit_application_type: true,
-            page: 1,
-            per_page: 20,
-            sort_by: 'status_with_assignment',
-            sort_direction: 'ascending',
-          }
-        end
-
+        let(:search_payload) { {} }
```
elken, line 139: Do these lines do anything?

ivanELEC: no actually - thought I needed them as a stand in for populating the stubs outside of context but that isn't the case - removed them

PR #678, opened 2024-08-09
CRM457-1794: caseworker CRM7 search and filter

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-XXX)

Screenshot 2024-08-09 at 15 10 13

https://crm457-1794-casewo-nscc-caseworker-dev.cloud-platform.service.justice.gov.uk/

### DIFF
--- a/app/forms/search_form.rb
+++ b/app/forms/search_form.rb
@@ -82,9 +82,7 @@ class SearchForm
       send(field).present?
     end
 
-    search_type = { crm7: 'claim', crm4: 'application' }
-
-    errors.add(:base, :nothing_specified, value: search_type[application_type.to_sym]) unless field_set
+    errors.add(:base, :nothing_specified, value: I18n.t("shared.submission_noun.#{application_type}")) unless field_set
   end
 
   def application_type_check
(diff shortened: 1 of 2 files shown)


app/forms/search_form.rb
```diff
@@ -82,7 +82,9 @@ def at_least_one_field_set
       send(field).present?
     end
 
-    errors.add(:base, :nothing_specified) unless field_set
+    search_type = { crm7: 'claim', crm4: 'application' }
```
patrick-laa, line 85: Nit: `'claim'` and `'application'` are technically user-facing strings, so shouldn't be hard-coded into our models. (If we were to offer our service in Welsh as well as English we'd have to refactor this). Really we should do something like:

```ruby
errors.add(:base, :nothing_specified, value: I18n.t("shared.submission_nouns.#{application_type}")) unless field_set
```

And then add something to config/locales/en.yml or similar

app/views/layouts/nsm/_primary_navigation.html.erb
```diff
@@ -0,0 +1,25 @@
+<% current ||= :your %>
```
patrick-laa, line 1: I think we can now delete app/views/layouts/_primary_navigation as you're replacing it with this new, better, appropriately namespaced one

patrick-laa (approved):

ivanELEC:
agree with Patrick's comments but otherwise happy with this - will approve once changes have been made

PR #692, opened 2024-08-15
CRM457-1860 CRM7 claim details show wrong info when rejected or granted and adjusted

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1860)

If the CW adjusts down and then GRANTS we only show the requested amount
If the CW adjusts up and then GRANTS we show the total of the requested and increased adjustment
If the CW Rejects it is always 0
If the CW Part grant it shows the requested amount +/ - any adjustment 

## Notes for reviewer
submission state now a variable in cost summary

### After changes:
rejected  

Screenshot 2024-08-15 at 16 52 53

granted (with adjustment)
Screenshot 2024-08-15 at 16 54 18


patrick-laa (approved):

app/models/claim.rb
```diff
@@ -62,6 +70,14 @@ def any_adjustments?
 
   private
 
+  def granted_and_allowed_less_than_claim
+    # TODO: refactor this to summed_costs.dig(:allowed_gross_cost, :text, :value)
```
patrick-laa, line 74: Nit: I don't like this `TODO`. If refactoring is the right thing to do, then either we should do the refactoring, or if it's too big a piece of work to do in this ticket, we should create another ticket and reference the number of that ticket in this comment, so that it's clear when and how the TODO is to be resolved.

spec/models/claims_spec.rb
```diff
@@ -30,4 +30,52 @@
...
+    end
+  end
+
+  describe '#formatted_allowed_total' do
+    context 'granted' do
+      it 'adjusted cost if adjusted more than claimed' do
+        claim = create(:claim, :increase_adjustment)
+        claim.state = Nsm::MakeDecisionForm::GRANTED
+        expect(claim).not_to receive(:formatted_claimed_total)
+        claim.formatted_allowed_total
```
patrick-laa, line 54: I think the tests should check that the output is what's expected here. E.g. at least that the allowed cost is greater than the claimed cost. And the next should check that the allowed cost is equal to the claimed cost. Otherwise this test just tests the implementation, not the behaviour.

PR #698, opened 2024-08-19
CRM457-1871: New UI for showing when work type changed

patrick-laa:
## Description of change
- Show all work types in work item summary
- Show asterisk in work item and core cost summary tables on row that no longer covers a work item because its work type has changed
- Add numbered footnotes to individual work item rows 

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1871)

### DIFF
--- a/app/javascript/lib/table-item-anchors.js
+++ b/app/javascript/lib/table-item-anchors.js
@@ -6,12 +6,8 @@ document.addEventListener("DOMContentLoaded", function () {
     });
     return btn;
   });
-  [...document.querySelectorAll("table[data-anchor-rows] tr[id] a")].forEach(
+  [...document.querySelectorAll("table[data-anchor-rows] tr[id] a:not([href^='#'])")].forEach(
     function (link) {
-      if (link.getAttribute('href')[0] === '#') {
-        // This is a link that does not go to a different page
-        return;
-      }
       link.addEventListener("click", function () {
         const nearestRow = link.closest("tr");
         if (nearestRow) {
--- a/app/javascript/lib/table-item-anchors.js
+++ b/app/javascript/lib/table-item-anchors.js
@@ -8,6 +8,10 @@ document.addEventListener("DOMContentLoaded", function () {
   });
   [...document.querySelectorAll("table[data-anchor-rows] tr[id] a")].forEach(
     function (link) {
+      if (link.getAttribute('href')[0] === '#') {
+        // This is a link that does not go to a different page
+        return;
+      }
       link.addEventListener("click", function () {
         const nearestRow = link.closest("tr");
         if (nearestRow) {
--- a/app/view_models/nsm/v1/core_cost_summary.rb
+++ b/app/view_models/nsm/v1/core_cost_summary.rb
@@ -1,6 +1,10 @@
 module Nsm
   module V1
     class CoreCostSummary < BaseViewModel
+      include ActionView::Helpers::UrlHelper
+      include ActionView::Helpers::TagHelper
+      include ActionView::Helpers::OutputSafetyHelper
+
       PROFIT_COSTS = 'profit_costs'.freeze
 
       attribute :submission
@@ -50,7 +54,7 @@ module Nsm
 
         @show_allowed ||=
           submission.part_grant? ||
-          [disbursements, letters_calls, *work_items.values].any? { |rows| rows.any?(&:changed?) }
+          [disbursements, letters_calls, work_items].any? { |rows| rows.any?(&:changed?) }
       end
 
       private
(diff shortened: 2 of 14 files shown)


app/javascript/lib/table-item-anchors.js
```diff
@@ -8,6 +8,10 @@ document.addEventListener("DOMContentLoaded", function () {
   });
   [...document.querySelectorAll("table[data-anchor-rows] tr[id] a")].forEach(
     function (link) {
+      if (link.getAttribute('href')[0] === '#') {
```
elken, line 11: Should probably adjust this check to be part of the `querySelector`, eg `a:not([href='#'])`

app/view_models/nsm/v1/core_cost_summary.rb
```diff
@@ -86,12 +90,22 @@ def calculate_disbursements(formatted:)
...
+        if show_allowed?
+          allowed_rows = work_items_of_type(type, type_type: :allowed)
+          allowed_net_cost = allowed_rows.sum(&:caseworker_amount) + extra_rows.sum(&:caseworker_amount)
+        end
+
+        calculate_hash(name, claimed_rows, allowed_rows, net_cost, allowed_net_cost, formatted)
+      end
+
+      def calculate_hash(name, claimed_rows, allowed_rows, net_cost, allowed_net_cost, formatted)
+        any_changed_type = show_allowed? && claimed_rows.any? { !_1.in?(allowed_rows) }
```
elken, line 105: Since you're only using `allowed_rows` here, wouldn't it be simpler to just calculate this bool outside and pass _that_ in? `claimed_rows` seems unused after too

app/view_models/nsm/v1/work_item_summary.rb
```diff
@@ -25,22 +26,23 @@ def header
...
+        WORK_TYPES.map do |work_type|
+          data = data_for(work_type)
           format_row(data)
         end
       end
 
       def footer
         format_row(
-          [t('total', numeric: false)] + summed_values(work_items, periods: false),
+          [t('total', numeric: false)] + summed_values(work_items, work_items, periods: false),
```
elken, line 37: Why is `work_items` being passed twice?

app/views/nsm/work_items/_adjusted.html.erb
```diff
@@ -22,19 +22,26 @@
...
-            <%=
-              if claim.editable_by?(current_user)
-                link_to(work_item.work_type, edit_nsm_claim_work_item_path(claim_id: claim.id, id: work_item.id), data: { turbo: 'false' })
-              else
-                link_to(work_item.work_type, nsm_claim_work_item_path(claim_id: claim.id, id: work_item.id), data: { turbo: 'false' })
-              end
-            %>
-            <% if work_item.work_type_original.present? %>
-              *
+            <% link = if claim.editable_by?(current_user)
```
elken, line 30: Would `polymorphic_path` be neater here?

app/views/nsm/work_items/_index.html.erb
```diff
@@ -37,19 +34,25 @@
...
-          <%=
-            if claim.editable_by?(current_user)
-              link_to(work_item.work_type, edit_nsm_claim_work_item_path(claim_id: claim.id, id: work_item.id), data: { turbo: 'false' })
-            else
-              link_to(work_item.work_type, nsm_claim_work_item_path(claim_id: claim.id, id: work_item.id), data: { turbo: 'false' })
-            end
-          %>
-          <% if work_item.work_type_original.present? %>
-            *
+          <% link = if claim.editable_by?(current_user)
```
elken, line 42: Would `polymorphic_path` be neater here?

elken (commented):
Few nits, nothing ground-breaking

app/javascript/lib/table-item-anchors.js
patrick-laa: Good shout, I forget how powerful CSS queries are, they can even handle "href _starts with_ a hash"

app/view_models/nsm/v1/core_cost_summary.rb
patrick-laa: Fair

app/view_models/nsm/v1/work_item_summary.rb
patrick-laa: It was because I changed the signature of `summed_values` to require both e "claimed work items" and a "allowed work items" - for the purposes of the footer, both are just "all work items". But it was clunky, so I've updated the signature to make the second one optional and default to the first one.

app/views/nsm/work_items/_adjusted.html.erb
patrick-laa: polymorphic_path doesn't quite do it in this case, simply because `work_item` is a `Nsm::V1::WorkItem` and without a bunch more tinkering that gets interpreted as `...nsm_v1_work_item...` which isn't a valid route. But `url_for` allows us to pass in an `action` option, so I've used that instead to avoid the if clause.

elken (approved):

PR #701, opened 2024-08-20
CRM457-1901: Adjust how disbursement adjustments are validated

elken:
## Description of change

This changes disbursement adjustments to only compare against the values submitted since the last update rather than the original values.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1901)

### DIFF
--- a/config/locales/en/nsm/work_items.yml
+++ b/config/locales/en/nsm/work_items.yml
@@ -127,7 +127,7 @@ en:
         nsm/work_item_form:
           attributes:
             base:
-              no_change: The new values cannot be the same as the values submitted by the provider
+              no_change: You cannot save changes as you have not changed anything. You can make a change and save it, or select cancel.
             uplift:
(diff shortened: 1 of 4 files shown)


config/locales/en/nsm/work_items.yml
```diff
@@ -125,7 +125,7 @@ en:
         nsm/work_item_form:
           attributes:
             base:
-              no_change: The new values cannot be the same as the values submitted by the provider
+              no_change: The new values cannot be the same the previously submitted values.
```
patrick-laa, line 128: We're going to need to do something similar for the four no_change strings in letters_and_calls.yml (there are four because there are 2 forms, and for each form we have the error message in 2 places depending on whether there is an uplift field or not)

elken: Ah I was searching for the wrong words, amending

PR #705, opened 2024-08-20
Don't initialize AWS if not needed

elken:
## Description of change

Startup time improvement since we don't always need to check for AWS

[Related discussion](https://mojdt.slack.com/archives/C05212WRK5J/p1724164271099309)


patrick-laa (approved):
Seems reasonable to me

PR #725, opened 2024-08-27
CRM457-1930: Resolve back button loop

elken:
## Description of change

Because of how the `history` was manipulating things, it was possible and easy to get stuck in an infinite loop of redirects.

It also broke a lot of other flows, so we drop the reliance on guesswork to just use fixed URLs with item anchors.

As noted in the commit, this might break some workflows that bypass the list pages, but the only real other option is to do some switching around what `Referrer` is or drop the "jump back to the row you came from" feature entirely.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1930)


app/javascript/lib/table-item-anchors.js
```diff
@@ -1,20 +0,0 @@
-document.addEventListener("DOMContentLoaded", function () {
-  [...document.querySelectorAll("a#back-button")].forEach(function (btn) {
-    btn.addEventListener("click", function (evt) {
-      evt.preventDefault();
-      history.back();
-    });
-    return btn;
-  });
-  [...document.querySelectorAll("table[data-anchor-rows] tr[id] a:not([href^='#'])")].forEach(
```
patrick-laa, line 9: I think we can remove the `data-anchor-rows` attributes from the various tables now.

app/views/nsm/disbursements/edit.html.erb
```diff
@@ -1,7 +1,7 @@
 <% title t('.page_title') %>
 
 <% content_for(:back_link) do %>
-  <%= link_to t('helpers.back_link'), "", id: "back-button", class: 'govuk-back-link', data: { turbo: 'false' } %>
+  <%= link_to t('helpers.back_link'), nsm_claim_disbursements_path(claim, anchor: item.id), class: 'govuk-back-link' %>
```
patrick-laa, line 4: This does behave weirdly if you were on adjusted_nsm_claim_disbursements_path and clicked through to this page from there, but unless we start storing state somehow to detect which table page you were most recently on, I think it's unavoidable, and this is still an improvement.

patrick-laa (approved):

app/javascript/lib/table-item-anchors.js
elken: Good point, will do

app/views/nsm/disbursements/edit.html.erb
elken: Well as I said in the PR body, we can do something similar to what `:back` does and use the referrer header to determine which URL to go back to.

I'll see how complex that ends up being as it may satiate the best of all worlds then

elken: Have gone for a much simpler method to just return the backlink path for a claim from an item which will also let us switch on potential other views in the future.

Think this is the best of all worlds now, unless there are any _other_ places that specifically link to a work item or disbursement.

app/view_models/nsm/v1/work_item.rb
```diff
@@ -136,6 +136,14 @@ def changed?
         adjustment_comment.present?
       end
 
+      def backlink_path(claim)
+        if any_adjustments?
+          Rails.application.routes.url_helpers.adjusted_nsm_claim_work_items_path(claim, anchor: id)
```
patrick-laa, line 141: The issue is that adjusted work items are visible as line items in both the 'Review and adjust' tab _and_ the 'Adjustments' tab, so it's incorrect to assume that the previous page for the edit screen of an adjusted work item is the 'Adjustments' tab.

elken: These are different pages now though, and I imagine it'd depend what you were doing last.

Should be one for UX to decide probably, I can see the case for both

patrick-laa (approved):
I think this'll do it. Still an edge case where if I visit a previously adjusted work item from the work items 'review and adjust' page, then try to make an edit but get a validation error, then click back, @referrer will be nil so we'll fall back to `backlink_path` which will send me to the adjusted work items page, but:

a) that's niche and
b) it's not a terrible UX - nothing breaks, and I'm only one click away from the tab I want to be on.

elken:
> I think this'll do it. Still an edge case where if I visit a previously adjusted work item from the work items 'review and adjust' page, then try to make an edit but get a validation error, then click back, @Referrer will be nil so we'll fall back to `backlink_path` which will send me to the adjusted work items page, but:
> 
> a) that's niche and b) it's not a terrible UX - nothing breaks, and I'm only one click away from the tab I want to be on.

Yeah that's the one I came up with too, probably a case of just deal with it if we ever hit it.

app/views/nsm/work_items/edit.html.erb
```diff
@@ -1,7 +1,7 @@
 <% title t('.page_title') %>
 
 <% content_for(:back_link) do %>
-  <%= link_to t('helpers.back_link'), "", id: "back-button", class: 'govuk-back-link', data: { turbo: 'false' } %>
+  <%= link_to t('helpers.back_link'), item.backlink_path(claim), class: 'govuk-back-link' %>
```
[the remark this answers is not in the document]
elken: Yes, we do 😓

PR #729, opened 2024-08-27
CRM457-1935: Sent back claims show with 'Closed' nav tab highlighted

patrick-laa:
## Description of change
Also excluded some big folders from rubocop which can significantly speed it up.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1935)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,6 +15,8 @@ AllCops:
     - 'lib/generators/**/*'
     - 'features/**/*'
     - 'vendor/**/*'
+    - 'tmp/**/*'
+    - 'node_modules/**/*'
 
 ####################################
 ## Customization for this project ##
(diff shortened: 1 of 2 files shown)


robinsk1 (approved):

elken:
While you're adding folders, I assume obvious `.` folder are excluded too? `.ruby-lsp` would be a good one if not

patrick-laa:
> While you're adding folders, I assume obvious `.` folders are excluded already `.ruby-lsp` would be a good one if not

Ooh, what's `.ruby-lsp`? I don't have that.

elken:
Ah it's the folder that https://github.com/Shopify/ruby-lsp generates to keep all its caches and dependencies.

image

Dawning on me just now I wonder if that's related to why the console is indeed so slow for me....

PR #737, opened 2024-08-29
CRM457-1946: Remove individual VAT figures

patrick-laa:
## Description of change
We shouldn't show individual vat figures for work items or letters/calls, as VAT is only applied once things are summed up.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1946)


elken (approved):
We love removing lots of code 😄

PR #738, opened 2024-08-29
CRM457-1940: Move cost adjustment asterisk outside of tab panel

elken:
## Description of change

Show the footnote denoting that records have changed type on the wider page rather than on the work items tab.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1940)

image


app/views/nsm/review_and_adjusts/show.html.erb
```diff
@@ -46,6 +46,40 @@
         <%= render partial: "nsm/#{scope}/index", locals: { claim:, records:, summary:, pagy: } %>
       </div>
     </div>
+
+    <%
+        type_changed_fields =
```
patrick-laa, line 51: You don't need this. Even if none of the work items that have changed type have changed from one core cost summary group to another (which is what this value indicates), we still want to show the `[*]` footnote, because there will be a `[*]` asterisk in the work items summary table. So the question is not `if type_changed_fields.any?` but rather `if type_changed_records.any?`

app/views/nsm/review_and_adjusts/show.html.erb
```diff
@@ -46,6 +46,40 @@
...
+            field.dig(:name, :text).include?(t("nsm.work_items.type_changes.asterisk"))
+          end
+    %>
+    <% if type_changed_fields.any? %>
+      <section>
+        <p id="fn*">
+          <%= t("nsm.work_items.type_changes.asterisk") %>
+          <%= t("nsm.work_items.type_changes.explanation") %>
+          <%= t("nsm.work_items.type_changes.see") %>
+          <% type_changed_records.each_with_index do |work_item, index| %>
```
patrick-laa, line 62: My understanding is that these footnotes _should_ still appear inside the tab - it's only the `[*]` footnote that's moving outside. Because the more specific footnotes area genuinely only relevant to the work items tab.

elken: Worth asking Mary then as I thought the oppposite 😛

patrick-laa: I completely misread the code, ignore me.

app/views/nsm/review_and_adjusts/show.html.erb
patrick-laa: I've resolved the other convo now that I've stopped misreading the code but I _think_ this one still stands - if you're on the work items page and a work item has changed type but hasn't changed to a different core cost row, `type_changed_fields` will be blank but we _do_ want to show the footnote.

elken: Yeah you're right I have this changed queued up.

patrick-laa (approved):

PR #757, opened 2024-09-10
CRM457-1973: Handle partial postcodes

patrick-laa:
## Description of change
- Maintain our own list of partial postcodes that straddle the M25. (Because it's a small and stable list)
- If a partial postcode is provided and it does _not_ fall in the above list, take a representative result from the API and check whether _that_ falls inside the M25.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1973)


app/services/location_service/outcodes_that_straddle_the_m25.yaml
imtiazAhmd, line 1: This looks exactly like Greater London Map. Do we need to have a list of all the partial postcodes of postal district? london.gov shows more than 52 listed here. https://www.london.gov.uk/sites/default/files/gla_postcode_map_a3_map1.pdf

patrick-laa: - We only need a list of postcodes that are both inside and outside the M25, i.e. the ones around the edges of the boundary, not the ones in the middle of London.
- Not all of the M25 is outside all of Greater London - they are two similar but distinct entities with overlapping boundaries, and trying to think about both will only lead to confusion!

app/services/location_service/outcodes_that_straddle_the_m25.yaml
```diff
@@ -0,0 +1,52 @@
+# Derived from https://www.google.com/maps/d/viewer?mid=1_yZ8LNwwjFuMMKHlThA4khMQDARD-DYl&z=11&ll=51.493316716269405%2C0.12035871093750039
```
robinsk1, line 1: Assuming these are a subset (m25 intersects) of the postcodes shown on that link, is the process of arriving at that subset documented?

robinsk1 (approved):

app/services/location_service/outcodes_that_straddle_the_m25.yaml
patrick-laa: Adding a comment

PR #763, opened 2024-09-13
CRM457-1949 delete adjustments

robinsk1:
## Description of change

delete single and all adjustments made by CW for CRM7 

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1949)

## Notes for reviewer
- an event `Event::UndoEdit` is created for each and every `delete` of an adjustment
- for a 'delete all adjustments' interaction each 'deletion' of an event records an `Event::Undo` and one single `Event::DeleteAdjustments` that is shown in claim history
- a 'deleted' adjustment is a `work_type`, `letter_and_call` or `disbursement` that is **reverted** to its original values
- `Services/AdjustmentDeleter` is now a base class that `PA` and `NSM` `AdjustmentDeleter` services inherit from
- redirects after a delete got to the type adjustment page listing adjustments unless there are no adjustments of any type and redirect will go to `claims/work_item`#index` page
### Before changes:
A CW is unable to delete adjustment changes

### After changes:
A CW can delete any adjustment they have made and costs/quantities will revert to original values

1. Log-in as CW and create some adjustments for existing work_items, calls_and_letters, or disbursements
2. Delete single adjustments 
3. Adjustment should no longer show and item should have reverted to original value

1. Log-in as CW and create some adjustments for existing work_items, calls_and_letters, or disbursements
2. Go to adjustments and click delete all adjustments
4. Check adjustments do not exist and items have revereted to original values
5. check that a delete all adjustments event is created with a comment in the claims history section

https://crm457-1949-delete-nscc-caseworker-dev.cloud-platform.service.justice.gov.uk/


app/services/nsm/all_adjustments_deleter.rb
```diff
@@ -0,0 +1,64 @@
+module Nsm
+  class AllAdjustmentsDeleter < AdjustmentDeleterBase
+    attr_reader :params, :current_user, :submission, :comment
+
+    def initialize(params, adjustment_type, current_user)
+      super
+      @comment = params[:nsm_delete_adjustments_form][:comment]
+    end
+
+    def call
```
patrick-laa, line 10: You might be able to simplify this. In PA we have a rule where when a submission is sent back we want to revert all adjustments, and we simply pull a clean copy of the blob from the app store and replace our adjusted blob with the clean one - https://github.com/ministryofjustice/laa-assess-crime-forms/blob/main/app/forms/prior_authority/send_back_form.rb#L57-L60. (I originally build something to explicitly revert each field like you've done, then David pointed out this as a simpler approach).

app/views/nsm/work_items/confirm_delete_adjustment.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= title t('.page_title') %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t('.page_title') %></h1>
+    <p>
+    <%= t('.cost_type', cost_type: @adjustment.work_type) %>
+    <br>
+    <%= t('.reason', reason: @adjustment.reason) %>
+    <br>
+    <%= raw t('.time_allowed', time_allowed: @adjustment.formatted_allowed_time_spent) %>
```
patrick-laa, line 10: _In general_ we want to avoid `raw` as it's overly permissive. If we change the string from `time_allowed` to `time_allowed_html` to use I18n.t's magic sanitisation logic, can we remove `raw`?

patrick-laa (approved):
I really like this code - the AdjustmentConcern is really neat and clear, the services are all very straightforward and easy to understand. Couple of suggestions but if you rejected both of them I'd still be happy with the code.

app/services/nsm/all_adjustments_deleter.rb
robinsk1: Cheers! one thing though wouldn't nuking all the adjustments like that obviate the item event tracking we get if we iterate through reverts? do we even need to track those events (we're doing it in PA, so i'm assuming that we need to do that for _some reason_)

patrick-laa: So if we did it that way we'd end up with a series of Edit events (the event history is outside the blob so wouldn't get nuked) and a single DeleteAdjustments event. We wouldn't have a set of individual UndoEdit events, but I kinda think that's reasonable - because there hasn't _been_ a series of undo events in the colloquial sense, there's just been one event which is the provider clicking the 'Delete all adjustments' button.

TBH what we use Event objects for has changed a lot. We used to treat them as things that could track very precisely the changes in state of an application over time, but these days we really only use them for the event history UI and occasional manual debugging/auditing, that's mostly replicated by the Kibana logs etc. I feel like the Edit/UndoEdit events aren't all that useful at this point so perhaps I'm more not-fussed about them than most.

robinsk1: ok, i was wondering if we were using those events indirectly to refresh what we had in app-store or other shenanigans but i guess not - i'll make that suggested change as it reduces a bit of clunk

jogilder (approved):
LGTM

PR #764, opened 2024-09-13
CRM457-1967: Stop reading from events

patrick-laa:
## Description of change
Always read from the blob rather than events
Also set the right timestamp for app_store_updated_at

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-1967)


app/view_models/prior_authority/v1/application_details.rb
```diff
@@ -74,18 +71,16 @@ def corrections_requested?
         updates_needed.include?(SendBackForm::INCORRECT_INFORMATION)
       end
 
-      delegate :latest_provider_update_event, to: :submission
-
       def provider_added_further_information?
-        latest_provider_update_event.details['comment'].present?
+        submission.data['further_information'].any? { _1['new'] }
```
patrick-laa, line 75: Todo: Migrations to ensure that `new` attributes are present.

jsugarman (approved):

app/view_models/prior_authority/v1/application_details.rb
jsugarman: Is that TODO to be part of this PR?

db/migrate/20240916092416_add_new_attribute_to_rfi_objects.rb
```diff
@@ -0,0 +1,25 @@
...
+        latest_fi = paa.data['further_information']&.max_by { _1['requested_at'] }
+        latest_fi && latest_fi['new'] = true
+      end
+
+      if added_correction
+        latest_correction = paa.data['incorrect_information']&.max_by { _1['requested_at'] }
+        latest_correction && latest_correction['new'] = true
+      end
+
+      paa.save!
```
jsugarman, line 22: i think we should probably have a `paa.save!(touch: false)` here, no?

PR #772, opened 2024-09-20
CRM457-2033:  Translations - MERGE THIS SECOND

patrick-laa:
## Description of change
* Pull translations from the [LAA Crime Forms Common](https://github.com/ministryofjustice/laa-crime-forms-common) gem instead of using the embedded one
* Allow translation fields either to be a hash with `en` and `value` keys, or a simple value string
* If changing work type, set the changed work type to be a value string rather than a hash (Note that a provider change will have to be merged before this can be merged so that changed work types in this new format can be synced back to provider)

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2033)


app/forms/base_adjustment_form.rb
```diff
@@ -48,10 +48,14 @@ def changed_value(val1, val2)
   def linked
     {
       type: self.class::LINKED_CLASS::LINKED_TYPE,
-      id: selected_record.dig(*self.class::LINKED_CLASS::ID_FIELDS),
+      id: linked_id(selected_record),
```
ivanELEC, line 51: just double checking - is this related to the translation changes or is this general code cleanup?

ivanELEC: never mind - understood the code after reading other parts

ivanELEC (approved):

PR #774, opened 2024-09-23
[CRM457-2054] only show delete adjustments for open claims

robinsk1:
## Description of change

remove delete links and delete all button for assesses claims 

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2054)

### DIFF
--- a/app/views/nsm/adjustments/show.html.erb
+++ b/app/views/nsm/adjustments/show.html.erb
@@ -9,7 +9,9 @@
         <%= t('.adjusted_costs') %>
       </h2>
 
-      <%= govuk_button_link_to( t('.delete_all_adjustments'), confirm_deletion_nsm_claim_adjustments_path(claim), secondary: true,  data: { turbo: "false" }) %>
+      <% unless claim.assessed? %>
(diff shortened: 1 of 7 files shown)


app/views/nsm/adjustments/show.html.erb
```diff
@@ -9,7 +9,9 @@
         <%= t('.adjusted_costs') %>
       </h2>
 
-      <%= govuk_button_link_to( t('.delete_all_adjustments'), confirm_deletion_nsm_claim_adjustments_path(claim), secondary: true,  data: { turbo: "false" }) %>
+      <% unless claim.assessed? %>
```
patrick-laa, line 12: I think this is the wrong check. In other places in these views we're only showing options to interact if `claim.editable_by?(current_user)`. We have a ticket to replace this rather ad-hoc authorisation with Pundit, but I think for now we do rely on views checking that the current user is allowed to edit a claim before showing editing options. (`editable_by?` checks things like "is the claim assigned to the current user?" as well as whether the claim is assessed or not).

patrick-laa (approved):

PR #777, opened 2024-09-24
CRM457-2002: Expire claims and show expired claims

patrick-laa:
## Description of change
- Add logic to expire claims once the deadline has passed
- Show expired claims with appropriate UI
- The above meant adding a new status, and tidging how we group statuses, because 'expired' claims haven't been assessed but we want to treat them the same as assessed claims in some ways.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2002)

### DIFF
--- a/config/locales/en/events.yml
+++ b/config/locales/en/events.yml
@@ -23,5 +23,7 @@ en:
     title:
       self: Caseworker removed self from claim
       secondary: Caseworker removed from claim by %{display_name}
+  event/delete_adjustments:
+    title: Caseworker deleted all adjustments
   event/expiry:
     title: Claim expired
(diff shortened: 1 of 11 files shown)


config/locales/en/events.yml
```diff
@@ -23,5 +23,5 @@ en:
     title:
       self: Caseworker removed self from claim
       secondary: Caseworker removed from claim by %{display_name}
-  event/delete_adjustments:
-    title: Caseworker deleted all adjustments
```
ivanELEC, line 27: Was this meant to be removed?

patrick-laa: Ooh no, that looks like a merge/rebase error. Will fix

PR #788, opened 2024-10-01
CRM457-2001 add CW RFI caseworker send back and claim detail UI

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2001)

## Notes for reviewer

This adds 

- a request for update element to the json blob via the send_back form
- UI updates to show a `provider updates` section to the claim details page as they are received from the provider

These changes are currently `hidden` behind a feature flag `nsm_rfi_loop` in UAT and Prod

Screenshot 2024-10-01 at 11 29 34
Screenshot 2024-10-01 at 11 29 51

We will probably need to set up 'test' env using this branch and corresponding changes in the provider app to test this feature


app/forms/nsm/send_back_form.rb
```diff
@@ -49,10 +49,12 @@ def add_further_information_data
       claim.data['resubmission_deadline'] = WorkingDayService.call(working_days_allowed)
       claim.data['further_information'] ||= []
       claim.data['further_information'] << {
+        documents: [],
         caseworker_id: current_user.id,
         information_requested: send_back_comment,
         requested_at: DateTime.current
       }
+      claim.data.merge!('updates_needed' => ['further_information'])
```
patrick-laa, line 57: This shouldn't be needed - it's only relevant for PA because in there there are two sorts of updates that are needed - further information and corrections. NSM has no concepts of correction, so we always know that any updates needed are _always_ further information.

app/views/nsm/claim_details/show.html.erb
```diff
@@ -12,6 +12,11 @@
         <p><strong><%= t(".sent_back", date: claim_summary.sent_back_on.to_fs(:stamp)) %></strong></p>
         <%= simple_format claim_summary.assessment_comment %>
       </div>
+    <% elsif claim.provider_updated? && FeatureFlags.nsm_rfi_loop.enabled? %>
```
patrick-laa, line 15: I think this feature flag check is redundant - a claim can't get to being `provider_updated` unless the RFI loop feature was enabled when it was sent back, and if it was enabled then, we don't want to break the UX for that claim if we subsequently switch the feature flag off.

app/forms/nsm/send_back_form.rb
jogilder: @patrick-laa, @robinsk1  is this not temporary? I thought that in the future we might have corrections as well as further info. I think maybe we check with UX and BA.

jogilder: @robinsk1 makes sense as I guess none of the other PRs have included it. Can you make sure you comment on the caseworker overall QA ticket explaining we haven't included that?

patrick-laa (approved):

ivanELEC (approved):

PR #792, opened 2024-10-03
HOTFIX for RFI nil documents 

robinsk1:
## Description of change

hotfix for when we might not have documents for further information - though this might actually just have been stray data in dev - an RFI send_back should always have an empty documents array that is populated by the provider on response if documents are added


patrick-laa (approved):
Suspect you'll need to add a test to get branch coverage up to speed

PR #793, opened 2024-10-03
[CRM457-2104] add service types to granted and part granted emails

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2104)

Update Prior Authority Notify granted and part-granted emails to provider to include the Service required and the Service provider details.

### DIFF
--- a/spec/mailers/prior_authority/feedback_messages/part_granted_feedback_spec.rb
+++ b/spec/mailers/prior_authority/feedback_messages/part_granted_feedback_spec.rb
@@ -45,6 +45,8 @@ RSpec.describe PriorAuthority::FeedbackMessages::PartGrantedFeedback do
         defendant_name: 'Abe Abrahams',
         application_total: '£300.00',
         part_grant_total: '£150.00',
+        service_required: 'Pathologist report',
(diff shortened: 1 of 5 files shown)


elken (approved):

spec/mailers/prior_authority/feedback_messages/part_granted_feedback_spec.rb
```diff
@@ -45,6 +45,8 @@
         defendant_name: 'Abe Abrahams',
         application_total: '£300.00',
         part_grant_total: '£150.00',
+        service_required: 'Pathologist report',
+        service_provider_details: 'ABC DEF<br>ABC, HIJ, SW1 1AA',
```
patrick-laa, line 49: I'm not _sure_ this is going to work. The [documentation](https://www.notifications.service.gov.uk/using-notify/formatting) says it uses markdown formatting so it might be that the <br> will be rendered as a literal string. But I may be way off the mark. Do we pass HTML tags to Notify anywhere else?

patrick-laa (approved):

PR #797, opened 2024-10-08
CRM457-2049: Fix RFI/'delete all adjustments' edge case

patrick-laa:
This reverts commit 892d599b033366fe755961c33916c9c4b211838d.

Now when we click 'delete all adjustments' we manually remove every adjustment from the payload instead of pulling a 'fresh copy' from the app store. This is because the app store version may contain adjustments if we have completed an RFI loop and adjustments were made before the claim was sent back.

Robin had written code to do it this way initially, and I (wrongly!) said it was much simpler to take the 'fresh copy' approach.

### DIFF
--- a/app/services/nsm/all_adjustments_deleter.rb
+++ b/app/services/nsm/all_adjustments_deleter.rb
@@ -32,8 +32,8 @@ module Nsm
       letters_and_calls.each do |letter_or_call|
         revert_fields.each do |field|
           revert(letter_or_call, field, letter_or_call['type']['value'])
-          letter_or_call.delete('adjustment_comment')
         end
+        letter_or_call.delete('adjustment_comment')
       end
     end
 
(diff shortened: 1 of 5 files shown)


app/services/nsm/all_adjustments_deleter.rb
```diff
@@ -10,12 +10,53 @@ def initialize(params, adjustment_type, current_user)
...
+    end
 
-      app_store_record = AppStoreClient.new.get_submission(submission)
-      submission.update!(data: app_store_record['application'])
+    def delete_letters_and_calls_adjustments
+      revert_fields = %w[uplift count]
+      letters_and_calls.each do |letter_or_call|
+        revert_fields.each do |field|
+          revert(letter_or_call, field, letter_or_call['type']['value'])
+          letter_or_call.delete('adjustment_comment')
```
elken, line 35: Why is this one inside the `revert_fields.each` but the other 2 are outside?

patrick-laa: Good spot! I don't think it should be here, so have moved.

elken (approved):
There's a fair amount of duplicated code here too, but it's nothing major.

PR #802, opened 2024-10-09
CRM457-2110: Change no data messaging

patrick-laa:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2110)


app/views/nsm/disbursements/_adjusted.html.erb
```diff
@@ -1,5 +1,7 @@
-<% if records.size.zero? %>
-  <p><%= t('.no_data') %></p>
+<% if claim.data['disbursements'].blank? %>
```
elken, line 1:
```ruby
if claim.data['disbursements'].blank? || records.empty?
```

elken (approved):

app/views/nsm/disbursements/_adjusted.html.erb
patrick-laa: In this context `records` are adjusted disbursements, and we _don't_ want to show `.no_disbursements` if `claim.data['disbursements']` is not blank but `records` is empty, as that implies there are disbursements but none of them are adjusted. So in that scenario we want to show `.no_adjustments`

elken: I misread completely and thought they both were `.no_disbursements` 🤣

PR #816, opened 2024-10-18
CRM457-2076: Go back to telling app store about new versions immediately

patrick-laa:
## Description of change
Go back to telling app store about new versions, and sync any unsynced new versions via a migration
[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2076)

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -24,7 +24,6 @@ class Event < ApplicationRecord
 
   LOCAL_EVENTS = [
     'Event::DraftDecision',
-    'Event::NewVersion',
     'Event::Edit',
     'Event::Note',
     'Event::UndoEdit',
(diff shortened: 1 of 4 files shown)


app/models/event.rb
```diff
@@ -24,7 +24,6 @@ class Event < ApplicationRecord
 
   LOCAL_EVENTS = [
     'Event::DraftDecision',
-    'Event::NewVersion',
```
robinsk1, line 27: Are we sure we want to delete this?

patrick-laa: That's not what `LOCAL_EVENTS` signifies. LOCAL_EVENTS is used to work out what the app store thinks the most recent updated_at date is, by knowing which events are synced to the app store.

elken: That makes sense, worth adding a comment?

elken (approved):
With that comment now I'm confident enough this is doing what it should then

PR #820, opened 2024-10-22
CRM457-2183: Allow downloading of RFI files

patrick-laa:
## Description of change
We were linking to the supporting evidence download path, but that only worked for documents listed as supporting evidence in the blob.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2183)

### DIFF
--- a/app/controllers/nsm/further_information_downloads_controller.rb
+++ b/app/controllers/nsm/further_information_downloads_controller.rb
@@ -1,0 +1,9 @@
+module Nsm
+  class FurtherInformationDownloadsController < Nsm::BaseController
+    include FileRedirectable
+
+    def show
+      redirect_to_file_download(params[:id], params[:file_name])
+    end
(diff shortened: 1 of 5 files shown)


elken (approved):
Looks to me like the raw file names are being put in, are we confident those will never be malicious?

elken:
> > Looks to me like the raw file names are being put in, are we confident those will never be malicious?
> 
> @elken it's a good point. I think we have to assume they _could_ be malicious. I can't off the top of my head see a way in which a crafted file name could do something we don't want, but that could very well be a failure of imagination on my part. I wonder if there's an exercise to be done to see if we can proof-of-concept something nasty happening with file names.

Might be worth ticketing up a spike for it, but you're _probably_ right that if it got into the system it's safe; I'd just personally sanitise on both sides with file names.

PR #827, opened 2024-10-29
CRM457-2198: Pricing

patrick-laa:
## Description of change
* Rather than do our own calculations that duplicate what the provider does, let them be done by the gem
* Rather than trust provider-submitted blob to find out the relevant rates, use the gem

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2198)

### DIFF
--- a/spec/mailers/nsm/feedback_messages/further_information_request_feeback_spec.rb
+++ b/spec/mailers/nsm/feedback_messages/further_information_request_feeback_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe Nsm::FeedbackMessages::FurtherInformationRequestFeedback do
   let(:ufn) { '123456/001' }
   let(:main_defendant_name) { 'Tracy Linklater' }
   let(:defendant_reference) { 'MAAT ID number: AB12123' }
-  let(:claim_total) { '£325.97' }
+  let(:claim_total) { '£359.76' }
   let(:date_to_respond_by) { 7.days.from_now.to_fs(:stamp) }
   let(:caseworker_information_requested) { 'Test Request' }
   let(:date) { DateTime.now.to_fs(:stamp) }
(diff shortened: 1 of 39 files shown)


spec/mailers/nsm/feedback_messages/further_information_request_feeback_spec.rb
```diff
@@ -12,7 +12,7 @@
   let(:ufn) { '123456/001' }
   let(:main_defendant_name) { 'Tracy Linklater' }
   let(:defendant_reference) { 'MAAT ID number: AB12123' }
-  let(:claim_total) { '£325.97' }
```
ivanELEC, line 15: Why did this value change?

ivanELEC (commented):
In general the implementation is clear enough but confused about the tests with the values changing and also -1 being added as a placeholder (I think we discussed this part in a meeting but I forgot the reasoning, is it just not needed in the test but needed for the gem's input?)

spec/mailers/nsm/feedback_messages/further_information_request_feeback_spec.rb
patrick-laa: In most of our tests our "pricing" values in the blob were unrealistic numbers, and we were using them to calculate the totals. Now we're not using values in the blob but rather actual pricing data from the gem, so the totals change.

patrick-laa:
>  is it just not needed in the test but needed for the gem's input?

@ivanELEC  So our tests need to use data that matches the schema, but now there are a bunch of "deprecated" fields in the schema - i.e. places where we have been putting things like pricing, even though now we don't want to be reading those values from the blob (and when we introduce a v2 schema we'll remove them entirely). By having technically-valid-but-absolutely-loopy values like `-1` in those fields in our test data, it allowed me to verify that we weren't naughtily continuing to use those values in the code - if we were, the tests would have blown up with negative figures.

elken (approved):

PR #837, opened 2024-11-07
CRM457-2216 pt 2: Talk to app store synchronously

patrick-laa:
## Description of change
As part of moving towards the caseworker app pushing/pulling to/from the app store directly rather than working with its local database and only occasionally syncing it, start making app store calls synchronous instead of async. This shouldn't have a noticeable performance impact, and reduces the number of edge cases where the 2 data stores can get out of sync.

Email sending is still async, as that is slow, so I've updated the flagging/locking mechanism to make sure email sending does not happen with stale data, and that failed jobs are recorded.

Also, because we already use the term "feedback" in the context of emails to handle emails triggered by filling in the feedback form, I've renamed the email stuff around sending emails to provider to remove the word "feedback", as this was confusing me and others.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2216)


robinsk1 (approved):
`perform_now` bypasses redis/sidekiq right? - all for that 😄

PR #845, opened 2024-11-13
[CMR457-2148] PA and NSM caseworker: amend 'make a decision' page

robinsk1:
## Description of change

Share design of make a decision page between PA and NSM claims

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2148)

e2e tests fixed here https://github.com/ministryofjustice/nsm-e2e-test/pull/168

### DIFF
--- a/app/views/nsm/make_decisions/edit.html.erb
+++ b/app/views/nsm/make_decisions/edit.html.erb
@@ -23,7 +23,7 @@
 
       <%= f.govuk_submit(t('.submit')) do %>
         <%= f.hidden_field :id %>
-        <%= govuk_button_link_to t('.save_and_exit'), nsm_claim_claim_details_path(claim), secondary: true, name: 'save_and_exit' %>
+        <%= f.govuk_submit t('.save_and_exit'),  secondary: true, name: 'save_and_exit' %>
       <% end %>
 
     <% end %>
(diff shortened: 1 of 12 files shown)


app/views/nsm/make_decisions/edit.html.erb
```diff
@@ -23,7 +23,7 @@
 
       <%= f.govuk_submit(t('.submit')) do %>
         <%= f.hidden_field :id %>
-        <%= link_to t('shared.cancel'), nsm_claim_claim_details_path(claim), class: 'govuk-link' %>
+        <%= govuk_button_link_to t('.save_and_exit'), nsm_claim_claim_details_path(claim), secondary: true, name: 'save_and_exit' %>
```
patrick-laa, line 26: Unfortunately I think there's some more work to do. If you're adding a 'Save and come back later' button to the UI, you need to implement 'save and come back later' functionality in the back end. If you look at `PriorAuthority::DecisionsController#create` you'll see if 'save_and_exit' is present it calls `stash` on the form, which shoves the values entered into the blob so that it can pre-populate the UI when you come back to it.

On the caseworker side this functionality isn't built-in, so you'll need to replicate it for NSM

patrick-laa (approved):

PR #846, opened 2024-11-14
CRM457-2154: add 'high value' filter to search

imtiazAhmd:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2154)


app/views/shared/_search.html.erb
```diff
@@ -65,6 +65,17 @@
           </fieldset>
         </div>
         <div>
+          <% if @search_form.application_type == 'crm7' %>
```
patrick-laa, line 68: This isn't quite the right check. On the dashboard, `application_type` is crm7 if the user has previously chosen the NSM radio button and submitted the form. So you'll get the weird behaviour that at first this field will be hidden, then if you choose NSM it will appear after you first submit the form. Whereas the ticket says that it should always be visible on the dashboard search form. You probably want `<% if choose_application_type || @search_form.application_type == 'crm7' %>`

app/forms/search_form.rb
```diff
@@ -26,6 +27,10 @@ def caseworkers
     [show_all] + User.order(:last_name, :first_name).map { Option.new(_1.id, _1.display_name) }
   end
 
+  def claim_values
+    [show_all] + %i[high_value].map { Option.new(_1, I18n.t("search.claim_values.#{_1}")) }
```
elken, line 31: Why is this mapping over an array with 1 value?

ivanELEC: image

it's not just mapping high_value, it's mapping high_value + anything in [show_all], the + operation happens before the map

elken: Oh high_value is a function, I thought it was a symbol

patrick-laa: @ivanELEC that doesn't show what you think it shows. Try `[1,2,3] + [4,5].map { _1 * 2 }`. If you were right you'd see `[2,4,6,8,10]`, but you'll actually see `[1,2,3,8,10]`. Ellis is right, the `map` is only being applied to `:high_value`, at which point you might as well just do `[show_all, Option.new(:high_value, I18n.t("search.claim_values.high_value")) }`

ivanELEC: thanks for keeping me honest there - looks like two wrongs don't make a right after all (even if it looks like they do) - will update to what you suggested as that's simpler to read

elken:
It has the do not merge label, doesn't need it in the title otherwise it'll accidentally go into the commit.

PR #861, opened 2024-12-02
CRM457-2303: Add "Additional fees" tabs

elken:
## Description of change

First batch of caseworker changes.

Due to time, I'll start review on this now while I add in the last coverage changes.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2303)

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -49,6 +49,21 @@ class Claim < Submission
     core_cost_summary.any_increased?
   end
 
+  def additional_fees
+    totals.fetch(:additional_fees, nil)
+  end
+
+  def youth_court_fee_elligible?
+    data['claim_type'] == 'non_standard_magistrate' &&
+      data['youth_court'] == 'yes' &&
+      data['plea_category'].match?(/category_(?:2|[12]a)$/)
+  end
+
+  def additional_fees?
+    FeatureFlags.youth_court_fee.enabled? &&
+      youth_court_fee_elligible?
+  end
+
   def adjustments_direction
     return :none unless any_cost_changes?
     return :mixed if any_cost_increases? && any_cost_reductions?
@@ -65,6 +80,7 @@ class Claim < Submission
     @rates ||= LaaCrimeFormsCommon::Pricing::Nsm.rates(data_for_calculation)
   end
 
+  # rubocop:disable Metrics/AbcSize
   def data_for_calculation
     {
       claim_type: BaseViewModel.build(:details_of_claim, self).claim_type.value,
@@ -74,8 +90,17 @@ class Claim < Submission
       work_items: work_items_for_calculation,
       disbursements: disbursements_for_calculation,
       letters_and_calls: letters_and_calls_for_calculation,
+      youth_court: data['youth_court'] == 'yes',
+      # TODO: CRM457-2306: Amend these as the fields will exist
+      claimed_youth_court_fee_included: data['include_youth_court_fee'],
+      plea_category: data['plea_category'],
+      # TODO: CRM457-2306: Remove these as the fields will exist
+      # :nocov:
+      assessed_youth_court_fee_included: data['youth_court_fee_adjustment_comment']&.present?,
+      # :nocov:
     }
   end
+  # rubocop:enable Metrics/AbcSize
 
   private
 
--- a/app/view_models/base_view_model.rb
+++ b/app/view_models/base_view_model.rb
@@ -34,7 +34,15 @@ class BaseViewModel
     end
 
     def process(&block)
-      result = rows.map(&block)
+      # :nocov:
+      # This change is needed because AdditionalFeesSummary doesn't
+      # call `build` the way other summary viewmodels do.
+      #
+      # We also can't easily cover it without writing a complex test
+      # for it, and since we want to hopefully remove all this
+      # evantually it's simpler to just not cover it.
+      result = rows&.map(&block)
+      # :nocov:
       return_array ? result : result[0]
     end
   end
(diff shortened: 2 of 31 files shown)


app/models/claim.rb
```diff
@@ -49,6 +49,15 @@ def any_cost_increases?
...
 
+  def additional_fees
+    totals.fetch(:additional_fees, nil)
+  end
+
+  def additional_fees?
+    FeatureFlags.youth_court_fee.enabled? &&
+      additional_fees&.dig(:total, :claimed_total_inc_vat)&.>(0)
+  end
+
```
ivanELEC, line 60: We should be checking whether claim is elligible for youth court fee (category *A, youth court == yes, nsm, on or after 6th dec)  instead of using total claimed amount

app/models/claim.rb
```diff
@@ -74,6 +83,11 @@ def data_for_calculation
       work_items: work_items_for_calculation,
       disbursements: disbursements_for_calculation,
       letters_and_calls: letters_and_calls_for_calculation,
+      youth_court: true,
+      claimed_youth_court_fee_included: data['include_youth_court_fee'],
+      plea_category: data['plea_category'],
```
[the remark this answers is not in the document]
ivanELEC: Use youth court fee adjustment comment to determine whether or not the caseworker has allowed the fee or not (if there is an adjusment comment - the caseworker is not allowing the fee, therefore this is false)

app/view_models/base_view_model.rb
```diff
@@ -34,7 +34,7 @@ def params(attributes)
     end
 
     def process(&block)
-      result = rows.map(&block)
+      result = rows&.map(&block)
```
ivanELEC, line 37: to test whether needed or not

app/view_models/nsm/v1/case_disposal.rb
```diff
@@ -3,6 +3,7 @@ module V1
     class CaseDisposal < BaseViewModel
       attribute :plea, :translated, scope: 'nsm.plea'
       attribute :plea_category, :translated, scope: 'nsm.plea_category'
+      attribute :include_youth_court_fee
```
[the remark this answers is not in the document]
ivanELEC: To confirm with Mary how this should appear i.e: 
Replace plea with custom text or have something like what provider has (Other: <custom text>)

config/locales/en/nsm/additional_fess.yml
```diff
@@ -0,0 +1,19 @@
+---
```
robinsk1, line 1: `additional_fess.yml` @elken you have something you want to tell us 😄 ?

elken: Not sure, could be a rebase gone awry

ivanELEC: good catch!

app/view_models/base_view_model.rb
elken: It is, probably due to how we set `@rows` _slightly_ different to the others.

app/models/claim.rb
```diff
@@ -74,8 +90,16 @@ def data_for_calculation
       work_items: work_items_for_calculation,
       disbursements: disbursements_for_calculation,
       letters_and_calls: letters_and_calls_for_calculation,
+      youth_court: data['youth_court'] == 'yes',
+      claimed_youth_court_fee_included: data['include_youth_court_fee'],
```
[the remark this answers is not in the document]
elken: Added

ivanELEC (approved):

robinsk1 (approved):

app/models/claim.rb
```diff
@@ -74,8 +90,17 @@ def data_for_calculation
       work_items: work_items_for_calculation,
       disbursements: disbursements_for_calculation,
       letters_and_calls: letters_and_calls_for_calculation,
+      youth_court: data['youth_court'] == 'yes',
+      # TODO: CRM457-2306: Amend these as the fields will exist
+      claimed_youth_court_fee_included: data['include_youth_court_fee'],
+      plea_category: data['plea_category'],
+      # TODO: CRM457-2306: Remove these as the fields will exist
+      # :nocov:
+      assessed_youth_court_fee_included: data['youth_court_fee_adjustment_comment']&.present?,
```
patrick-laa, line 99: This is backwards. If there is an adjustment comment it will be because a caseworker has _removed_ the youth court fee.

The value of assessed_youth_court_fee_included is always available in `data['youth_court_fee_included']` so you can just use that.

app/models/claim.rb
```diff
@@ -74,8 +90,17 @@ def data_for_calculation
       work_items: work_items_for_calculation,
       disbursements: disbursements_for_calculation,
       letters_and_calls: letters_and_calls_for_calculation,
+      youth_court: data['youth_court'] == 'yes',
+      # TODO: CRM457-2306: Amend these as the fields will exist
+      claimed_youth_court_fee_included: data['include_youth_court_fee'],
```
patrick-laa, line 95: Strictly, unless the adjustment flow goes waaaay off piste, this should be `data.fetch('include_youth_court_fee_original', data['include_youth_court_fee'])` - i.e. if there is an "_original" value use that, otherwise use the default value

app/models/claim.rb
elken: Both of these will be picked up in 2306, but yeah you're right

ivanELEC: Just to clarify the intention is to put the correct input for this field as a part of the next PR (that handles adjustments) - this is essentially acting as a placeholder as youth_court_adjustment_comment can't exist yet. We probably should have entered just "false" for now, with a comment explaining what we intend to do

elken: It can't be assessed yet though right? That's #2?

PR #865, opened 2024-12-05
Ship youth court fee

elken:
## Description of change

Deploy the changes for YCF

PR #867, opened 2024-12-05
CRM457-2303: Caseworker part 1 fixes

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2303)

## Notes for reviewer

- Take changes to additional_fees? check from enable PR (so we can test before enabling)
- Fix/add to existing additional_fees controller and system tests
- Fix logic on Additional Fee view model for assessed claims
- Split claim and data factories and fix associated tests

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -102,7 +102,7 @@ class Claim < Submission
   private
 
   def youth_court_fee_claimed
-    data.fetch('include_youth_court_fee_original', data['include_youth_court_fee'])
+    data['include_youth_court_fee_original'] || data['include_youth_court_fee']
   end
 
   def granted_and_allowed_less_than_claim
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -81,7 +81,6 @@ class Claim < Submission
     @rates ||= LaaCrimeFormsCommon::Pricing::Nsm.rates(data_for_calculation)
   end
 
-  # rubocop:disable Metrics/AbcSize
   def data_for_calculation
     {
       claim_type: BaseViewModel.build(:details_of_claim, self).claim_type.value,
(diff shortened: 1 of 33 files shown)


app/models/claim.rb
```diff
@@ -91,19 +91,24 @@ def data_for_calculation
...
-      assessed_youth_court_fee_included: data['youth_court_fee_adjustment_comment']&.present?,
+      assessed_youth_court_fee_included: data['include_youth_court_fee'],
       # :nocov:
     }
   end
-  # rubocop:enable Metrics/AbcSize
 
   private
 
+  def youth_court_fee_allowed
```
patrick-laa, line 104: I think this should be called `youth_court_fee_claimed`

app/view_models/nsm/v1/additional_fee.rb
```diff
@@ -39,7 +39,7 @@ def table_fields
         [
           I18n.t("nsm.additional_fees.index.#{type}"),
           format(claimed_total_exc_vat),
-          format(assessed_total_inc_vat.positive? ? assessed_total_inc_vat : nil),
+          format(any_adjustments? ? assessed_total_exc_vat : nil),
```
patrick-laa, line 42: Possibly worth adding a comment pointing out that for `any_adjustments?` to work we will have to set the `adjustment_comment` attribute on this class when instantiating it in `AdditionalFeesSummary`

patrick-laa (approved):

app/models/claim.rb
elken: I also preferred your data.fetch call

ivanELEC: will update on both counts

elken:
Just a heads up it looks like some of these fixes clash with what's on https://github.com/ministryofjustice/laa-assess-crime-forms/pull/865 so that will have to be rebased

app/models/claim.rb
```diff
@@ -102,7 +102,7 @@ def data_for_calculation
   private
 
   def youth_court_fee_claimed
-    data.fetch('include_youth_court_fee_original', data['include_youth_court_fee'])
+    data['include_youth_court_fee_original'] || data['include_youth_court_fee']
```
elken, line 105: Be aware that if this is false, it'll check the other value. I don't think it's what you want to do.

ivanELEC: true (which means somewhere also this is untested - will fix this to be more specific, tbh I'm pretty sure rubocop auto-changed this from code where I check specifically if data['include_youth_court_fee_original'] is nil using .nil?

patrick-laa: So this is only a problem if `include_youth_court_fee_original` is false but `include_youth_court_fee` is true, i.e. if the provider didn't claim the fee but the caseworker awarded it to them anyway. _At the moment_ this can't bite us because I don't think we intend to give caseworkers functionality do that, but it would be a kindness to our future selves to do this properly just in case requirements are added.

elken: This is being handled in https://github.com/ministryofjustice/laa-assess-crime-forms/pull/865 (for our future selves)

PR #869, opened 2024-12-09
[CRM457-2306] NSM Caseworker: Adjust youth court fee

robinsk1:
## Description of change

Caseworkers needs to be able to review the new additional fee information and make appropriate an assessment.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2306)

### DIFF
--- a/app/controllers/nsm/additional_fees_controller.rb
+++ b/app/controllers/nsm/additional_fees_controller.rb
@@ -7,7 +7,7 @@ module Nsm
     include Nsm::AdjustmentConcern
 
     FORMS = {
-      'youth_court_fee' => AdditionalFeesForm::YouthCourtFee
+      'youth_court_fee' => YouthCourtFeeForm
     }.freeze
 
     def index
@@ -43,8 +43,25 @@ module Nsm
         model.type == params[:id].to_sym
       end
 
-      render locals: { claim:, item:, }
+      form = form_class.new(claim:, item:, **item.form_attributes)
+      render :edit, locals: { claim:, item:, form: }
     end
+
+    # rubocop:disable Metrics/AbcSize
+    def update
+      authorize(claim, :edit?)
+      rows = BaseViewModel.build(:additional_fees_summary, claim).rows
+      item = rows.detect do |model|
+        model.type == params[:id].to_sym
+      end
+      form = form_class.new(claim:, item:, **form_params)
+      if form.save!
+        redirect_to nsm_claim_additional_fees_path(claim)
+      else
+        render :edit, locals: { claim:, item:, form: }
+      end
+    end
+    # rubocop:enable Metrics/AbcSize
 
     def adjusted
       authorize(claim, :show?)
@@ -69,6 +86,12 @@ module Nsm
       FORMS[params[:id]]
     end
     # :nocov:
+
+    def form_params
+      params.require(:nsm_youth_court_fee_form)
+            .permit(:remove_youth_court_fee, :explanation)
+            .merge(current_user:)
+    end
 
     def fail_if_no_additional_fees
       raise ActionController::RoutingError, 'Not Found' unless claim.additional_fees?
--- a/app/forms/nsm/youth_court_fee_form.rb
+++ b/app/forms/nsm/youth_court_fee_form.rb
@@ -1,0 +1,34 @@
+module Nsm
+  class YouthCourtFeeForm < BaseAdjustmentForm
+    COMMENT_FIELD = 'youth_court_fee_adjustment_comment'.freeze
+    LINKED_CLASS = V1::AdditionalFee
+
+    attribute :remove_youth_court_fee, :boolean
+    attribute :youth_court_fee_adjustment_comment
+
+    # UX only requires and explanation for when CW wants to remove fee
+    validates :explanation, presence: true, if: :remove_youth_court_fee
+
+    def save
+      return false unless valid?
+
+      process_field(value: !remove_youth_court_fee, field: 'include_youth_court_fee')
+
+      true
+    end
+
+    private
+
+    def no_change_field
+      :remove_youth_court_fee
+    end
+
+    def selected_record
+      @selected_record ||= submission.data
+    end
+
+    def data_has_changed?
+      remove_youth_court_fee == item.include_youth_court_fee
+    end
+  end
+end
--- a/app/view_models/nsm/v1/additional_fee.rb
+++ b/app/view_models/nsm/v1/additional_fee.rb
@@ -5,7 +5,6 @@ module Nsm
 
       attribute :type
       attribute :submission
-      attribute :include_youth_court_fee
       attribute :claimed_total_exc_vat
       attribute :claimed_vatable
       attribute :assessed_total_exc_vat
@@ -24,6 +23,14 @@ module Nsm
           ]
         end
 
+        def adjusted_headers
+          [
+            t('.fee_type', width: 'govuk-!-width-one-fifth', numeric: false),
+            t('.reason_for_adjustments', numeric: false),
+            t('.net_cost_allowed')
+          ]
+        end
+
         private
 
         def t(key, width: nil, numeric: true, scope: 'nsm.additional_fees.index')
@@ -32,6 +39,14 @@ module Nsm
             numeric: numeric,
             width: width
           }
+        end
+      end
+
+      def backlink_path(claim)
+        if any_adjustments?
+          Rails.application.routes.url_helpers.adjusted_nsm_claim_additional_fees_path(claim)
+        else
+          Rails.application.routes.url_helpers.nsm_claim_additional_fees_path(claim)
         end
       end
 
@@ -48,17 +63,6 @@ module Nsm
           '.additional_fee' => I18n.t("nsm.additional_fees.edit.#{type}"),
           '.net_cost_claimed' => NumberTo.pounds(claimed_total_exc_vat)
         }
-      end
-
-      def backlink_path(claim)
-        # :nocov:
-        # TODO: CRM457-2306: Remove these as the fields will exist
-        if any_adjustments?
-          Rails.application.routes.url_helpers.adjusted_nsm_claim_additional_fees_path(claim)
-        # :nocov:
-        else
-          Rails.application.routes.url_helpers.nsm_claim_additional_fees_path(claim)
-        end
       end
 
       private
--- a/app/view_models/nsm/v1/core_cost_summary.rb
+++ b/app/view_models/nsm/v1/core_cost_summary.rb
@@ -51,15 +51,15 @@ module Nsm
       end
 
       def any_changed?
-        [disbursements, letters_calls, work_items].any? { |rows| rows.any?(&:changed?) }
+        [disbursements, letters_calls, work_items, additional_fees].any? { |rows| rows.any?(&:changed?) }
       end
 
       def any_reduced?
-        [disbursements, letters_calls, work_items].any? { |rows| rows.any?(&:reduced?) }
+        [disbursements, letters_calls, work_items, additional_fees].any? { |rows| rows.any?(&:reduced?) }
       end
 
       def any_increased?
-        [disbursements, letters_calls, work_items].any? { |rows| rows.any?(&:increased?) }
+        [disbursements, letters_calls, work_items, additional_fees].any? { |rows| rows.any?(&:increased?) }
       end
 
       private
@@ -121,6 +121,10 @@ module Nsm
 
       def disbursements
         @disbursements ||= BaseViewModel.build(:disbursement, submission, 'disbursements')
+      end
+
+      def additional_fees
+        @additional_fees ||= AdditionalFeesSummary.new('submission' => submission).rows
       end
     end
   end
--- a/app/view_models/nsm/v1/youth_court_fee.rb
+++ b/app/view_models/nsm/v1/youth_court_fee.rb
@@ -1,0 +1,85 @@
+module Nsm
+  module V1
+    class YouthCourtFee < AdditionalFee
+      adjustable_attribute :include_youth_court_fee, :boolean
+      attribute :youth_court_fee_adjustment_comment, :string
+
+      def key
+        'summary_table'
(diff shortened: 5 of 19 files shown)


app/forms/nsm/youth_court_fee_form.rb
```diff
@@ -0,0 +1,48 @@
...
+    def explanation_required?
+      remove_youth_court_fee == 'true'
+    end
+
+    def selected_record
+      @selected_record ||= submission.data
+    end
+
+    def data_has_changed?
+      remove_youth_court_fee != item.include_youth_court_fee
```
patrick-laa, line 39: This looks suspect, for 2 reasons:

First, I believe `remove_youth_court_fee` is a string, so comparing it with a boolean will go all kinda wokny.

Second, if it was a boolean,  if there was a youth court fee, and I wanted to remove it, then we'd be evaluating `true != true` which will return false, so this would suggest that the data has _not_ changed, even though I was trying to change the data.

app/forms/nsm/youth_court_fee_form.rb
```diff
@@ -0,0 +1,48 @@
...
+      process_field(value: !remove_bool, field: 'include_youth_court_fee')
+
+      true
+    end
+
+    private
+
+    def ensure_original_field_value_set(field)
+      if selected_record["#{field}_original"].present?
+        selected_record.delete("#{field}_original")
```
patrick-laa, line 23: I _think_ I see what you're trying to do here - you're trying to make it so that if a caseworker removes the youth court fee and then re-edits the form to add it back in again, we revert the state to its original state without any `_original` stuff, right? But because you're shoehorning that logic into a side effect of a method called `ensure_original_field_value_set` and what you're doing is _removing_ the original field, it makes for pretty hairy reading. 

I would suggest _not_ doing this - the system should be resilient enough to cope with there being a now-redundant "_original" field, because it copes with it for every other adjustment form, and it will spare the devs bunch of confusion here.

app/controllers/nsm/additional_fees_controller.rb
```diff
@@ -38,12 +38,21 @@ def show
 
     def edit
       authorize(claim, :edit?)
-      rows = BaseViewModel.build(:additional_fees_summary, claim).rows
-      item = rows.detect do |model|
-        model.type == params[:id].to_sym
-      end
 
-      render locals: { claim:, item:, }
+      item = BaseViewModel.build(:additional_fee, claim)
```
elken, line 42: There will likely be other kinds of fees in the future so we can't rely on this always being the right one hence the `detect` check to get the right one

app/forms/nsm/youth_court_fee_form.rb
```diff
@@ -0,0 +1,48 @@
...
+    COMMENT_FIELD = 'youth_court_fee_adjustment_comment'.freeze
+    LINKED_CLASS = V1::AdditionalFee
+
+    attribute :remove_youth_court_fee
+    attribute :youth_court_fee_adjustment_comment
+
+    def save
+      return false unless valid?
+
+      remove_bool = ActiveModel::Type::Boolean.new.cast(remove_youth_court_fee)
```
elken, line 12: Isn't this just `!explanation_required?` since that checks if it's equal to true?

app/forms/nsm/youth_court_fee_form.rb
```diff
@@ -0,0 +1,48 @@
...
+      if selected_record["#{field}_original"].present?
+        selected_record.delete("#{field}_original")
+        selected_record.delete(self.class::COMMENT_FIELD)
+      else
+        selected_record["#{field}_original"] ||= selected_record[field]
+      end
+    end
+
+    def explanation_required?
+      remove_youth_court_fee == 'true'
```
patrick-laa, line 31: I'm not convinced by this. In every other form we require an explanation any time the record is changed from its previous state, even if it's being reverted to its original state. That might be the system behaving wrong, but if it is wrong we should fix it everywhere. If you really want to buck the trend here and only here, it'd be worth adding a comment explaining why this is a special case with a different UX.

app/view_models/nsm/v1/core_cost_summary.rb
```diff
@@ -122,6 +122,10 @@ def letters_calls
       def disbursements
         @disbursements ||= BaseViewModel.build(:disbursement, submission, 'disbursements')
       end
+
+      def additional_fees
+        @additional_fees ||= BaseViewModel.build(:additional_fee, submission)
```
elken, line 127: Might be worth adding dummy extra `AdditionalFees` to check how they affect areas like this so it doesn't bite us later

app/views/nsm/additional_fees/_adjusted.html.erb
```diff
@@ -1,6 +1,28 @@
...
   <div class="govuk-grid-column-full">
-    <p><%= t('.no_data') %></p>
+    <% if records.rows.none? { _1.any_adjustments? } %>
+      <p><%= t('.no_data') %></p>
+    <% else %>
+      <%=
+      caption = t('.table_caption')
+      head = Nsm::V1::AdditionalFee.headers
+      rows = records.rows.reject{|item| item.type == :total}.map do |additional_fee|
+        fields = additional_fee.table_fields.dup
```
elken, line 10: I don't know if this needs a check to only show rows that actually have adjustments

elken (changes requested):

app/view_models/nsm/v1/additional_fee.rb
```diff
@@ -15,6 +15,9 @@ class AdditionalFee < BaseWithAdjustments
       attribute :claimed_total_inc_vat
       attribute :assessed_total_inc_vat
 
+      adjustable_attribute :include_youth_court_fee, :boolean
+      attribute :youth_court_fee_adjustment_comment, :string
```
patrick-laa, line 19: So as I understand it, the goal hitherto has been to make this model agnostic about what _type_ of additional fee is being claimed, to make it more future-proof. Is there any way we can keep `youth_court`-specificness out  of this file?

app/forms/nsm/youth_court_fee_form.rb
patrick-laa, line 12: Could you not do `attribute :remove_youth_court_fee, :boolean` on line 6 and save yourself having to cast it manually here and do string comparisons on line 31?

app/forms/nsm/youth_court_fee_form.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    def selected_record
+      @selected_record ||= submission.data
+    end
+
+    def data_has_changed?
+      remove_youth_court_fee != item.include_youth_court_fee
+    end
+
+    def linked
```
patrick-laa, line 42: How come you need to override this?

app/view_models/nsm/v1/core_cost_summary.rb
```diff
@@ -51,7 +51,7 @@ def show_allowed?
       end
 
       def any_changed?
-        [disbursements, letters_calls, work_items].any? { |rows| rows.any?(&:changed?) }
+        [disbursements, letters_calls, work_items].any? { |rows| rows.any?(&:changed?) } || additional_fees.changed?
```
patrick-laa, line 54: Again, this breaks the supposition that there could be multiple additional fees, by checking a single AdditionalFee model that knows it's a youth court fee. Slightly nicer if we could add an `additional_fees` array to the array on the left hand side of this line.

app/forms/nsm/youth_court_fee_form.rb
ivanELEC: do we even need to override this method since explanation is overidden?

app/view_models/nsm/v1/additional_fee.rb
ivanELEC: guess we could have a youth_court_fee view model that inherits this? then update everything else in line with the idea that we will have multiple different types of additional fee

robinsk1: refactored youth_court_fee model to inherit from additional_fee

app/view_models/nsm/v1/core_cost_summary.rb
robinsk1: `additional_fees` now an array of fee objects built via AdditionalFeeSummary object #rows method

elken: Sorry I wasn't clear; by that I mean adding test cases to show what having multiple fees looks like.

app/views/nsm/additional_fees/_adjusted.html.erb
robinsk1: done

app/view_models/nsm/v1/youth_court_fee.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        return false if include_ycf.nil? || include_ycf_original.nil?
+
+        include_ycf != include_ycf_original
+      end
+      alias changed? any_adjustments?
+
+      def form_attributes
+        remove_bool = ActiveModel::Type::Boolean.new.cast(include_youth_court_fee)
+        remove_youth_court_fee = include_youth_court_fee_original.nil? ? nil : !remove_bool
```
ivanELEC, line 44: where does the variable/method include_youth_court_fee_original come from here?

robinsk1: from the form as text so we need to cast it

app/views/nsm/additional_fees/_adjusted.html.erb
```diff
@@ -1,6 +1,30 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-    <p><%= t('.no_data') %></p>
+    <% if records.rows.none? { _1.any_adjustments? } %>
+      <p><%= t('.no_data') %></p>
+    <% else %>
+      <%=
+      caption = t('.table_caption')
+      head = Nsm::V1::AdditionalFee.headers
+      rows = records.rows.reject{|item| item.type == :total}.map do |additional_fee|
```
ivanELEC, line 9: nit: can use _1 shorthand notation here if you want: 
record.rows.reject{ _1.type == :total } 

but not too fussed either way

app/view_models/nsm/v1/youth_court_fee.rb
```diff
@@ -0,0 +1,57 @@
...
+        include_ycf_original = submission.data['include_youth_court_fee_original']
+
+        return false if include_ycf.nil? || include_ycf_original.nil?
+
+        include_ycf != include_ycf_original
+      end
+      alias changed? any_adjustments?
+
+      def form_attributes
+        remove_bool = ActiveModel::Type::Boolean.new.cast(include_youth_court_fee)
```
patrick-laa, line 43: You shouldn't need to cast this as a boolean - it's already defined as one on line 4 here

app/view_models/nsm/v1/additional_fee.rb
```diff
@@ -35,6 +35,16 @@ def t(key, width: nil, numeric: true, scope: 'nsm.additional_fees.index')
...
+      def backlink_path(claim)
+        # :nocov:
+        if any_adjustments?
+          Rails.application.routes.url_helpers.adjusted_nsm_claim_additional_fees_path(claim)
+        # :nocov:
+        else
+          Rails.application.routes.url_helpers.nsm_claim_additional_fees_path(claim)
+        end
+      end
+
```
ivanELEC, line 47: is the include_youth_court_fee attribute in this class still needed as it's captured in youth_court_fee as an adjustable attribute?

app/view_models/nsm/v1/youth_court_fee.rb
elken: The form doesn't pull `_original` is what Ivan's raising

patrick-laa: On line 4 we've defined `include_youth_court_fee` as an `adjustable_attribute`, which means we get `include_youth_court_fee_original` as an attribute automatically.

ivanELEC: yeah - I assume this would need to come from the submission

elken: Ah see https://github.com/ministryofjustice/laa-assess-crime-forms/pull/869#discussion_r1877930989, this is resolved

app/view_models/nsm/v1/additional_fee.rb
robinsk1: removed attribute include_youth_court_fee as duplicated in child model, renamed additional_fee_spec to youth_court_fee_spec

app/view_models/nsm/v1/youth_court_fee.rb
robinsk1: done

app/view_models/nsm/v1/core_cost_summary.rb
robinsk1: added a dummy view model/dummy additional fee to test this in AdditionalFeeSummary spec

elken (commented):
Couple more

app/view_models/nsm/v1/additional_fee.rb
```diff
@@ -35,6 +34,16 @@ def t(key, width: nil, numeric: true, scope: 'nsm.additional_fees.index')
         end
       end
 
+      def backlink_path(claim)
+        # :nocov:
```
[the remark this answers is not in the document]
robinsk1: done

app/forms/nsm/youth_court_fee_form.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    private
+
+    def selected_record
+      @selected_record ||= submission.data
+    end
+
+    def data_has_changed?
+      !remove_youth_court_fee != item.include_youth_court_fee
```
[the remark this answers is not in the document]
robinsk1: yeah 😅

elken (approved):

ivanELEC (approved):

PR #871, opened 2024-12-10
CRM457-2300: Delete youth court adjustments

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2300)

### Before changes:
image

### After changes:
image

image

### DIFF
--- a/app/controllers/concerns/nsm/adjustment_concern.rb
+++ b/app/controllers/concerns/nsm/adjustment_concern.rb
@@ -2,8 +2,17 @@ module Nsm
   module AdjustmentConcern
     extend ActiveSupport::Concern
     def confirm_deletion
+      raise 'Cannot delete non-existent adjustment' unless any_adjustments?
+
       authorize(claim, :update?)
-      @adjustment = adjustments.find { _1.id == params[:id] }
+      @adjustment = if additional_fee?
+                      view_model
+                    else
+                      view_model.filter(&:any_adjustments?).find do |item|
+                        item.id == params[:id]
+                      end
+                    end
+
       render :confirm_delete_adjustment, locals: { claim_id: params[:claim_id], id: params[:id] }
     end
 
@@ -19,20 +28,31 @@ module Nsm
       claim.any_adjustments? ? { action: :adjusted } : nsm_claim_work_items_path
     end
 
-    def adjustments
-      @adjustments ||= BaseViewModel
-                       .build(resource_klass, claim, json_search_field)
-                       .filter(&:any_adjustments?)
+    def view_model
+      @view_model ||= BaseViewModel.build(resource_klass, claim, nesting)
+    end
+
+    def nesting
+      additional_fee? ? nil : json_search_field
+    end
+
+    def additional_fee?
+      json_search_field == 'additional_fees'
     end
 
     def resource_klass
       return :letter_and_call if json_search_field == 'letters_and_calls'
+      return params[:id].to_sym if additional_fee?
 
       @resource_klass ||= controller_name.singularize.to_sym
     end
 
     def json_search_field
       @json_search_field ||= controller_name
+    end
+
+    def any_adjustments?
+      additional_fee? ? view_model.any_adjustments? : view_model.any?(&:any_adjustments?)
     end
   end
 end
--- a/app/services/nsm/all_adjustments_deleter.rb
+++ b/app/services/nsm/all_adjustments_deleter.rb
@@ -11,6 +11,7 @@ module Nsm
       delete_work_item_adjustments if work_items
       delete_letters_and_calls_adjustments if letters_and_calls
       delete_disbursement_adjustments if disbursements
+      delete_youth_court_fee_adjustment if youth_court_fee_adjustment_comment
       ::Event::DeleteAdjustments.build(submission:, comment:, current_user:)
     end
 
@@ -42,6 +43,15 @@ module Nsm
         end
         disbursement.delete('adjustment_comment')
       end
+    end
+
+    def delete_youth_court_fee_adjustment
+      revert(submission.data, 'include_youth_court_fee', 'additional_fees')
+      submission.data.delete('youth_court_fee_adjustment_comment')
+    end
+
+    def youth_court_fee_adjustment_comment
+      submission.data['youth_court_fee_adjustment_comment']
     end
 
     def letters_and_calls
--- a/app/view_models/nsm/v1/youth_court_fee.rb
+++ b/app/view_models/nsm/v1/youth_court_fee.rb
@@ -48,8 +48,8 @@ module Nsm
 
       def caseworker_fields
         {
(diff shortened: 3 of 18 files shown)


app/controllers/concerns/nsm/adjustment_concern.rb
```diff
@@ -1,11 +1,21 @@
...
+    # rubocop:disable Style/NumberedParameters
     def confirm_deletion
       authorize(claim, :update?)
-      @adjustment = adjustments.find { _1.id == params[:id] }
+
+      @adjustment = if additional_fee?
+                      view_model
+                    else
+                      view_model.filter(&:any_adjustments?).find do
+                        _1.id == params[:id]
```
elken, line 12: I think Rubocop is right here, it's not much cleaner to use a numbered parameter

app/view_models/nsm/v1/youth_court_fee.rb
```diff
@@ -73,6 +73,18 @@ def calculation
...
 
+      def type_name
+        I18n.t(".nsm.youth_court_fee_adjustments.#{key}.youth_court_fee")
+      end
+
+      def allowed_net
+        NumberTo.pounds(caseworker_amount)
+      end
+
+      def reason
```
elken, line 84: Might as well use this in `caseworker_fields`

spec/services/nsm/all_adjustments_deleter_spec.rb
```diff
@@ -143,6 +153,12 @@
         expect(subject).not_to receive(:delete_disbursement_adjustments)
         subject.call
       end
+
+      it 'wont try to youth court fee adjustment if none' do
```
elken, line 157: nit: "wont try to remove"

spec/system/nsm/additional_fees_spec.rb
```diff
@@ -229,6 +226,21 @@
         click_button 'Save changes'
         expect(page).not_to have_content('Adjustments')
       end
+
+      it 'can remove youth court fee adjustment through the adjustments tab' do
+        visit adjusted_nsm_claim_additional_fees_path(claim)
+
+        click_on 'Additional fees'
+
+        # might need to have a more specific selector
```
elken, line 235: How do we do this for letters and calls?

Something like (napkin Ruby)

```ruby
within(:xpath, "//tr[.//th[contains(text(),'Youth Court Fee')]]") do
  click_button("Delete")
end
```

Should work I think if there's no pre-existing standard

elken (commented):
Nothing earth-shattering here, imagine this won't be as long as the other CW tickets

spec/system/nsm/additional_fees_spec.rb
patrick-laa: Elsewhere in these system specs (e.g. disbursements_spec.rb) we use the pattern `within('.govuk-table__row', text: 'Youth court fee') do` which is marginally easier to read

app/controllers/concerns/nsm/adjustment_concern.rb
ivanELEC: true - I think when I initially ignored this it was all in one line without the do block

spec/services/nsm/all_adjustments_deleter_spec.rb
ivanELEC: another instance of "don't copy and paste" aha

app/services/nsm/all_adjustments_deleter.rb
```diff
@@ -44,6 +45,15 @@ def delete_disbursement_adjustments
       end
     end
 
+    def delete_youth_court_fee_adjustment
+      revert(submission.data, 'include_youth_court_fee', 'additional_fees')
+      submission.data.delete('youth_court_fee_adjustment_comment')
+    end
+
+    def youth_court_fee
```
patrick-laa, line 53: Nit: This method is called `youth_court_fee` but doesn't return a youth court fee, it returns whether or not the youth court fee has been adjusted. (The other methods below it are badly named too, but less so, e.g. `letters_and_calls` returns letters and calls that have been adjusted, but at least they are letters and calls). Maybe we could call this `youth_court_fee_adjustment_comment`

Equally, because it's only used once, we don't need to memoize it.

spec/system/nsm/additional_fees_spec.rb
ivanELEC: It just isn't tested in the letters_and_calls integration test (only the remove all uplifts button is)

ivanELEC: will try this though!

elken: That's fine then, Patrick's version is much cleaner

app/views/nsm/additional_fees/_adjusted.html.erb
```diff
@@ -6,16 +6,23 @@
...
         next unless additional_fee.any_adjustments?
 
         fields = additional_fee.adjusted_fields.dup
-        destination = if policy(claim).update?
+        destination = if editable
                         edit_nsm_claim_additional_fee_path(claim_id: claim.id, id: additional_fee.type)
                       else
                         nsm_claim_additional_fee_path(claim_id: claim.id, id: additional_fee.type)
                       end
+        deleteable = confirm_deletion_nsm_claim_additional_fee_path(claim_id: claim.id, id: additional_fee.type) if editable
```
patrick-laa, line 23: Nit: This reads a bit confusingly. `deleteable` sounds like it would be a boolean, but it's actually the deletion path which might be nil. Perhaps simpler would be:

```
deletion_path = confirm_deletion_nsm_claim_additional_fee_path(claim_id: claim.id, id: additional_fee.type) 
...
fields << link_to(t('.delete'), deletion_path, data: { turbo: 'false' }) if editable
```

spec/system/nsm/additional_fees_spec.rb
ivanELEC: ok - was having a lot of fun with xpath helper chrome extensions aha

app/controllers/concerns/nsm/adjustment_concern.rb
```diff
@@ -1,11 +1,21 @@
 module Nsm
   module AdjustmentConcern
     extend ActiveSupport::Concern
+    # rubocop:disable Style/NumberedParameters
     def confirm_deletion
       authorize(claim, :update?)
-      @adjustment = adjustments.find { _1.id == params[:id] }
+
+      @adjustment = if additional_fee?
+                      view_model
```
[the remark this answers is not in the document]
elken: What happens if you try and submit a deletion for an adjustment where the ID doesn't exist?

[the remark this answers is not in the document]
ivanELEC: [update from call] shown in meeting that this triggers a generic error page (will be considered as part of more hollistic messaging convo with UR)

robinsk1 (approved):

PR #877, opened 2024-12-16
CRM457-2217: Simplify use of events

patrick-laa:
## Description of change
- Get rid of Postgis as we don't need it any more and it's holding us back from upgrading to Rails 8
- Stop handling 'Edit' events, because events are _purely_ there to populate the history UI, and we don't show edits in the history UI (this means we can ditch the whole LINKED_ID system)
- Stop marking some events as 'public' and some as 'history' - ALL events are history events, NONE are public

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2217)


elken (approved):
I never got what `LINKED_CLASS` was for anyway

robinsk1 (approved):

PR #905, opened 2025-01-29
CRM457-2405 Adds feedback url 

imtiazAhmd:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2405)


app/helpers/application_helper.rb
```diff
@@ -92,4 +92,25 @@ def infer_claim_section
       :open
     end
   end
+
+  def current_namespace
```
robinsk1, line 96: you can probably squash method this into feedback_url and get the same result

elken (approved):
Ah I thought the main layout was already setup to handle this, makes sense to me 😄

PR #908, opened 2025-02-05
[CRM457-2427] update GDPR messaging

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2427)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -88,7 +88,7 @@ en:
           last_updated: Last updated
           status_with_assignment: Status
     gdpr_documents_deleted:
-      message: Uploaded files deleted. Your uploads are deleted after 6 months to keep your data safe.
+      message: Uploaded files deleted. Uploads are deleted after 6 months to keep the provider’s data safe.
   generic:
     cost: cost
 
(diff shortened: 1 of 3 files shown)


elken (commented):
I'd prefer to use `I18n.t` in tests instead of hard-coding the message

robinsk1:
> I'd prefer to use `I18n.t` in tests instead of hard-coding the message

i guess in this case, as its a system test its testing the expected output of `I18n.t` so it sort of makes sense why we have all similar test conditions like that set as strings

elken:
> > I'd prefer to use `I18n.t` in tests instead of hard-coding the message
> 
> i guess in this case, as its a system test its testing the expected output of `I18n.t` so it sort of makes sense why we have all similar test conditions like that set as strings

And I disagree with it for exactly the reason this PR shows :P The output changing content shouldn't constitute a test failure, you're just testing if the expected message appears.

robinsk1:
> > > I'd prefer to use `I18n.t` in tests instead of hard-coding the message
> > 
> > 
> > i guess in this case, as its a system test its testing the expected output of `I18n.t` so it sort of makes sense why we have all similar test conditions like that set as strings
> 
> And I disagree with it for exactly the reason this PR shows :P The output changing content shouldn't constitute a test failure, you're just testing if the expected message appears.

yeah i know its brittle, maybe the rationale was to test the value of translations at one point and we took that approach as theres a few others in there that follow that pattern. Changed this one though 😜

PR #945, opened 2025-04-14
Add a user management UI

elken:
## Description of change

Add a new UI to allow users to be managed outside of the now-deprecated rake task.

Also restricts which services caseworkers should be able to use.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2535)

## Notes for reviewer

This will need adjustments to how every other dev has their users
setup locally. If you're using the default seeded users, you won't
have to make any changes other than re-running `bundle exec db:seed`
again.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,0 +1,50 @@
+class UsersController < ApplicationController
+  before_action :authorize_supervisor
+  before_action :set_default_table_sort_options, only: %i[show]
+
+  DEFAULT_SORT = 'name'.freeze
+
+  def show
+    pagy, records = pagy_array(users)
+
+    render :index, locals: { pagy: pagy, users: records }
+  end
+
+  private
+
+  def users
+    direction = @sort_direction == 'descending' ? :desc : :asc
+    case @sort_by
+    when 'name'
+      User.order(first_name: direction, last_name: direction)
+    when 'role'
+      User.joins(:roles)
+          .select('users.*, roles.role_type')
+          .distinct
+          .order("roles.role_type #{direction}")
+    else
+      User.order(params[:sort_by] => direction)
+    end
+  end
+
+  def authorize_supervisor
+    authorize :dashboard, :show?
+  end
+
+  def controller_params
+    params.permit(
+      :sort_by,
+      :sort_direction,
+      :page
+    )
+  end
+
+  def param_validator
+    @param_validator ||= UsersParams.new(controller_params)
+  end
+
+  def set_default_table_sort_options
+    @sort_by = controller_params.fetch(:sort_by, DEFAULT_SORT)
+    @sort_direction = controller_params.fetch(:sort_direction, 'descending')
+  end
+end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,41 +1,84 @@
 class UsersController < ApplicationController
   before_action :authorize_supervisor
-  before_action :set_default_table_sort_options, only: %i[show]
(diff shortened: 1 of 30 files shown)


elken:
Marked as draft for now until the live migration code is ready, but this is ready for review/QA

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+    render :index, locals: { pagy:, users: }
+  end
+
+  def new
+    @form_object = UserForm.new
+  end
+
+  def edit
+    @user = User.find(controller_params['id'])
+    user = @user.attributes.slice('first_name', 'last_name', 'email')
```
ivanELEC, line 19: nitpick: not sure if there's existing reccomendations for this type of thing but having the instance variable and local variable be the same name could get confusing (i.e. people might think the local variable = the instance variable when diagnosing. That said, there isn't really a better name for the variable so not really sure what the best way is to go forward, could call it form_user or something but that also is awkward. Maybe we just leave it as is (still going to leave this comment in case you have any ideas)

app/forms/user_form.rb
```diff
@@ -0,0 +1,69 @@
...
+      first_name: first_name,
+      last_name: last_name,
+      email: email,
+      auth_oid: SecureRandom.uuid,
+      roles: [Role.new(role_type:, service:)]
+    )
+
+    return true if user.errors.empty?
+
+    user.errors.map { |error| errors.add(error.attribute.to_sym, error.type.to_sym) }
```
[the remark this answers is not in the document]
ivanELEC: so for example: if we were to enforce auth_oid to be unique (we don't but actually maybe we should), if SecureRandom.uuid somehow generated the same id twice, this would get captured in the form errors. Would we want the user to know this ?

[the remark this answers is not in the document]
ivanELEC: I wonder if the form validation is enough and the user validation can stay as a redirection to the "Sorry something went wrong" page

app/models/user.rb
```diff
@@ -13,7 +13,7 @@ class User < ApplicationRecord
   scope :pending_activation, -> { where(auth_subject_id: nil, deactivated_at: nil) }
 
   def display_name
-    "#{first_name} #{last_name}"
+    "#{first_name.capitalize} #{last_name.capitalize}"
```
ivanELEC, line 16: This is potentially one we put out to the team if we haven't already - there are instances where surnames aren't capitalised french stuff like e.g. de Soya and the only [context of when it should be capitalised is the start of a sentence](https://english.stackexchange.com/questions/246157/should-i-capitalize-a-persons-last-name-if-their-name-starts-with-a-lower-case) and in this context, it's not a sentence just displaying the name. Since the supervisors now have control over the names, and not us I think it's ok to leave that up to them to keep tidy (in instances where it's not capitalised in the db but it should be)

app/controllers/users_controller.rb
elken: Don't think there's really much better as you say

app/models/user.rb
elken: Yeah raise it to the team, I'm not convinced it's a huge deal

app/forms/user_form.rb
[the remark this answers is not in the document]
ivanELEC: Ok that's good with me and I understand why we are bringing this in via the User model and not just in the form itself (unnecessary to hand roll that email uniqueness code) - I do think this warrants a comment on the User model saying that any constraints added will affect the User management forms

ivanELEC (approved):

ivanELEC (approved):

PR #963, opened 2025-05-13
Remove N+1 queries when loading user info

elken:
## Description of change

Policies against users run on every page and always go through a user's roles a few times. It only makes sense to compute this once per object lifetime.

[Link to relevant discussion](https://mojdt.slack.com/archives/C06UR7V28RZ/p1746802375906659)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,16 +12,24 @@ class User < ApplicationRecord
   scope :active, -> { where(deactivated_at: nil).where.not(auth_subject_id: nil) }
   scope :pending_activation, -> { where(auth_subject_id: nil, deactivated_at: nil) }
 
+  def nsm_access?
+    @nsm_access ||= roles.nsm_access.exists?
+  end
+
+  def pa_access?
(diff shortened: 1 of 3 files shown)


elken:
Dev deploy failed probably because of the branch name

ivanELEC (approved):

ivanELEC:
I think the "+" in the branch name might be messing something up  in helm possibly

app/models/user.rb
```diff
@@ -12,16 +12,24 @@ class User < ApplicationRecord
   scope :active, -> { where(deactivated_at: nil).where.not(auth_subject_id: nil) }
   scope :pending_activation, -> { where(auth_subject_id: nil, deactivated_at: nil) }
 
+  def nsm_access?
+    @nsm_access ||= roles.nsm_access.exists?
```
robinsk1, line 16: does the memoized var and method name not need to be the same? maybe not

elken: Only by convention

PR #995, opened 2025-06-18
Add flatware to parallize test suite

elken:
## Description of change

The current test suite takes several minutes each invocation, and
makes keeping track of coverage a pain as you have to re-run the
entire suite.

Now, the suite runs in tens of seconds.

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2622)


ivanELEC (commented):
Great stuff!!! Could we get a little bit added to the readme indicating how people can run tests locally (as they have to run the `bundle exec flatware fan rails db:test:prepare` command beforehand right?)

ivanELEC (approved):

PR #1039, opened 2025-09-04
Do not merge: Testing emails don't send on submission update

ivanELEC:


elken:
I assume this is done @ivanELEC ?

PR #1055, opened 2025-09-29
Create multi-step NSM payment request form

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2711)

### DIFF
--- a/app/javascript/component/payments/total_costs.js
+++ b/app/javascript/component/payments/total_costs.js
@@ -30,7 +30,7 @@ function init() {
 
     if (profit && disbursement && travel && waiting) {
       const sum = profit.plus(disbursement).plus(travel).plus(waiting).toFixed(2);
-      total.textContent = sum;
+      total.textContent = `£${sum}`;
     } else {
       total.textContent = "—";
     }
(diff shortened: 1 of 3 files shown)


app/forms/payments/steps/nsm/claim_detail_form.rb
```diff
@@ -0,0 +1,36 @@
...
+        attribute :date_completed, :date
+
+        validates :defendant_first_name, :defendant_last_name,
+                  :hearing_outcome_code, :matter_type, :court,
+                  presence: true
+
+        validates :number_of_defendants, :number_of_attendances, presence: true, is_a_number: true,
+          numericality: { only_integer: true, greater_than_or_equal_to: 0 }
+
+        validates :youth_court, inclusion: { in: [true, false] }
```
elken, line 26: It's already a boolean, is this needed?

elken (commented):
Lots of abstractions being introduced here, think we want to start moving away not introducing more. 

Also seems to be lots of fiddling with session and form data directly, would avoid that unless it's an absolute requirement for some reason

elken (approved):

PR #1184, opened 2026-04-01
Modsec remediations

TawneeOwl:
## Description of change

Amends modsec ruleset to increase security


RobBoothMOJ:
Was created as part of remediation work for ModSec issue 05/2026.   We agreed the way forward to be a targeted adjustment on a form-by-form basis.

PR #1198, opened 2026-04-17
Onboarding Cleanup

ivanELEC:
## Description of change

[Link to relevant ticket](https://trello.com/c/MW0CDvgK/178-tracking-updates-needed-for-our-onboarding-process)

## Notes for reviewer

- Removes postgis from brewfile as that isn't used in this service (only used in app store)
- Updates postgresql version in brewfile 
- Removes custom seed tasks and updates readme (those tasks were only relevant when we had claim data in assess and not app store)


Brewfile
```diff
@@ -1,6 +1,5 @@
 tap 'homebrew/cask'
-brew 'postgresql@14'
+brew 'postgresql@18'
```
robinsk1: is this matching  version we use in prod?

robinsk1 (dismissed):

Brewfile
ivanELEC: good point - we are on v17 in prod not v18, thought for now maybe it makes sense to keep as v18 here in preparation to moving to v18 in prod sooner rather than later?

robinsk1 (approved):

PR #1203, opened 2026-04-22
CRM457-3035 clear out stale payment request session

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3035)

## Notes for reviewer
- fixes scenarios where a user may abandon a payment request creation leaving orphaned session data 
- pro-active session clean-up

### DIFF
--- a/app/services/decisions/multi_step_form_session.rb
+++ b/app/services/decisions/multi_step_form_session.rb
@@ -37,7 +37,7 @@ module Decisions
     end
 
     def no_existing_ref?
-      answers['laa_reference'].blank? && answers['linked_nsm_ref'].blank?
+      answers['laa_reference'].blank? && answers['linked_nsm_ref'].blank? && answers['nsm_claim_id'].blank?
     end
 
     def ac_claim_details_incomplete?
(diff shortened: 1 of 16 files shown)


ivanELEC (dismissed):

app/services/decisions/multi_step_form_session.rb
```diff
@@ -17,9 +17,11 @@ def answers=(claim_hash)
       data['answers'] = claim_hash
     end
 
-    def reset_answers
+    def reset_answers(persist_type: false)
+      request_type = self['request_type']
       data['answers'] = { 'id' => id,
         'idempotency_token' => SecureRandom.uuid }
+      data['answers']['request_type'] = request_type if persist_type
```
robinsk1: i'm not sure about this - only a new record will require this and a new record will only be of a specific type so its sort of moot - it also dilutes the idea of this method which is to solely reset answers. Can we not achieve the same result by intentionally making the call to new record the thing that sets the type, deletes any existing session data? that seems more intentional that controlling that behaviour from claim-search controller

ivanELEC (approved):

PR #1206, opened 2026-04-23
CRM457-3017: Clear Payments Data When Creating from New Record

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3017)

## Notes for reviewer
Fixing bug caught in payments journeys where data wasn't cleared out when users navigates to claim search page 
Clears all answer data apart from request_type when user navigates back to claim search page

### DIFF
--- a/app/controllers/payments/steps/office_code_search_controller.rb
+++ b/app/controllers/payments/steps/office_code_search_controller.rb
@@ -1,22 +1,12 @@
 module Payments
   module Steps
     class OfficeCodeSearchController < BaseController
-      before_action :clear_session_payment, only: [:edit], if: -> { params[:new_record].present? }
-
       def edit
         @form_object = Payments::Steps::OfficeCodeSearchForm.build(multi_step_form_session.answers, multi_step_form_session:)
       end
 
       def update
         update_and_advance(Payments::Steps::OfficeCodeSearchForm, as: :office_code_search)
-      end
-
-      private
-
-      def clear_session_payment
-        request_type = multi_step_form_session['request_type']
-        multi_step_form_session.reset_answers
-        multi_step_form_session['request_type'] = request_type
       end
     end
   end
--- a/app/controllers/payments/steps/office_code_search_controller.rb
+++ b/app/controllers/payments/steps/office_code_search_controller.rb
@@ -1,12 +1,22 @@
 module Payments
   module Steps
     class OfficeCodeSearchController < BaseController
+      before_action :clear_session_payment, only: [:edit], if: -> { params[:new_record].present? }
+
       def edit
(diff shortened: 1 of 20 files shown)


robinsk1 (dismissed):

app/controllers/payments/steps/office_code_search_controller.rb
```diff
@@ -1,13 +1,23 @@
...
         @form_object = Payments::Steps::OfficeCodeSearchForm.build(multi_step_form_session.answers, multi_step_form_session:)
       end
 
       def update
         update_and_advance(Payments::Steps::OfficeCodeSearchForm, as: :office_code_search)
       end
+
+      private
+
+      def clear_session_payment
```
cmrgov: Tiny nitpick but this is triggered by the edit `GET` request which clears payment session data, which potentially makes the reset URL unsafe to refresh, revisit from browser history, or prefetch. 

Might be safer to use a POST action

app/views/payments/steps/claim_search/edit.html.erb
```diff
@@ -17,7 +17,7 @@
 </div>
 
 <div class="govuk-!-margin-top-3">
-  <%= govuk_button_to t('.new_record'), edit_payments_steps_office_code_search_path, method: :get, secondary: true %>
+  <%= govuk_link_to t('.new_record'), edit_payments_steps_office_code_search_path(new_record: 'true'), class: "govuk-button govuk-button--secondary", method: :get, secondary: true %>
```
cmrgov: Can we use `govuk_button_link_to` here instead of `govuk_link_to` with manual button classes? This repo is on `govuk-components` 6.1.0, which supports that helper, and it keeps the GOV.UK button-link markup/classes consistent.

https://govuk-components.x-govuk.org/helpers/button/#link-styled-like-a-button

cmrgov (commented):
LGTM @ivanELEC  -  just a couple of questions about using GET to clear session and govcomponents button link

app/views/payments/steps/claim_search/edit.html.erb
ivanELEC: yeah good shout thanks

app/controllers/payments/steps/office_code_search_controller.rb
ivanELEC: We intentionally want the data to regresh on refresh and revisit from browser history but yeah pre-fetch might be a problem and yeah semantically we aren't just navigating to the page, we are executing data functions and then navigating to the page so POST makes more sense - will look into it

ivanELEC: I've made those changes now - used PUT instead of a POST because
1. we aren't creating a new record - we are updating existing info 
2. means we don't have to create an entire new route and method

cmrgov (approved):

PR #1228, opened 2026-05-12
CRM457-3018: Reduce duplicate App Store calls

cmrgov:
## Description of change

[CRM457-3018](https://dsdmoj.atlassian.net/browse/CRM457-3018)

Reduces duplicate App Store calls when rendering the Assess NSM claim details page.

## Notes for reviewer

- Calculates payment request eligibility once in `Nsm::ClaimDetailsController#show`.
- Reuses that local in the claim details view instead of calling `eligible_for_payment_request?` twice.
- Updates `ApplicationHelper#infer_claim_section` to reuse the already-loaded claim where available.
- Uses `assigned_user_id` for the “your claims” check to avoid an extra user lookup.
- Does not change Application Store APIs, query shape, or payloads.

1. Open an assessed or granted NSM claim with payments enabled.
2. Confirm the “Create payment request” button still appears.
3. Confirm App Store calls are reduced to:
   - one `GET /v1/submissions/:id`
   - one `POST /v1/payment_requests/searches`

## Testing

- `CI=true bundle exec rspec spec/controllers/nsm/claim_details_controller_spec.rb`
- `CI=true bundle exec rspec spec/system/nsm/overview_spec.rb spec/system/payments/nsm_payment_request_spec.rb spec/system/nsm/assessment_spec.rb`
- `bundle exec rubocop`
- `bundle exec erb_lint app/views/nsm/claim_details/show.html.erb`

Full local RSpec had an unrelated deterministic failure in `spec/system/users_spec.rb:45` after this branch was rebased. The focused CRM457-3018 coverage and relevant NSM/payment flows pass.

[CRM457-3018]: https://dsdmoj.atlassian.net/browse/CRM457-3018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/nsm/claim_details_controller.rb
+++ b/app/controllers/nsm/claim_details_controller.rb
@@ -4,8 +4,9 @@ module Nsm
       authorize(claim)
       claim_summary = BaseViewModel.build(:claim_summary, claim)
       claim_details = ClaimDetails::Table.new(claim)
+      payment_request_eligible = claim.assessed? && claim.eligible_for_payment_request?
 
-      render locals: { claim:, claim_summary:, claim_details:, provider_updates: }
+      render locals: { claim:, claim_summary:, claim_details:, provider_updates:, payment_request_eligible: }
     end
 
     private
--- a/app/controllers/nsm/claim_details_controller.rb
+++ b/app/controllers/nsm/claim_details_controller.rb
@@ -4,9 +4,8 @@ module Nsm
       authorize(claim)
       claim_summary = BaseViewModel.build(:claim_summary, claim)
       claim_details = ClaimDetails::Table.new(claim)
-      payment_request_eligible = claim.assessed? && claim.eligible_for_payment_request?
 
-      render locals: { claim:, claim_summary:, claim_details:, provider_updates:, payment_request_eligible: }
+      render locals: { claim:, claim_summary:, claim_details:, provider_updates: }
     end
 
     private
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -80,17 +80,26 @@ module ApplicationHelper
   end
 
   def infer_claim_section
-    claim_id = params.fetch(:claim_id, params.fetch(:id, nil))
-    return unless claim_id
+    current_claim = current_claim_for_section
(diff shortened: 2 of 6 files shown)


app/helpers/application_helper.rb
```diff
@@ -80,19 +80,28 @@ def gbp_field_value(value)
...
+    elsif current_claim.assigned_user_id == current_user.id
       :your
     else
       :open
     end
   end
 
+  def current_claim_for_section
+    claim_id = params.fetch(:claim_id, params.fetch(:id, nil))
+    return unless claim_id
```
ivanELEC, line 97: potentially out of scope but should this raise a specific error if claim_id is null (instead of following through with claim_id = nil and failing at a later point?

ivanELEC (dismissed):

app/controllers/nsm/claim_details_controller.rb
```diff
@@ -4,8 +4,9 @@ def show
       authorize(claim)
       claim_summary = BaseViewModel.build(:claim_summary, claim)
       claim_details = ClaimDetails::Table.new(claim)
+      payment_request_eligible = claim.assessed? && claim.eligible_for_payment_request?
 
-      render locals: { claim:, claim_summary:, claim_details:, provider_updates: }
+      render locals: { claim:, claim_summary:, claim_details:, provider_updates:, payment_request_eligible: }
```
robinsk1: setting a flag here seems excessive - could you not just do this is by memoizing what `claim.eligible_for_payment_request?` returns?

cmrgov: Good shout - memoizing is tidier. Needed to do an unconventional memoization to handle the false properly though

app/helpers/application_helper.rb
cmrgov: I’d leave this as-is for now. The nav helper can run on pages without a claim param, where nil falls back to `:open`; claim-specific pages already validate `claim_id` in the controller.

ivanELEC (approved):

PR #1246, opened 2026-06-01
CRM457-3078: Assess: Provider Data API schedule endpoint cannot retreive data in UAT

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3078)

## Notes for reviewer

- Add `effectiveDate` param to schedule endpoint usage
- Add tests

### DIFF
--- a/app/services/provider_data/provider_data_api_client.rb
+++ b/app/services/provider_data/provider_data_api_client.rb
@@ -6,6 +6,8 @@ module ProviderData
     headers 'X-Authorization' => ENV.fetch('PROVIDER_API_KEY')
     format :json
 
+    PROVIDER_API_EFFECTIVE_DATE_PARAM = ENV.fetch('PROVIDER_API_EFFECTIVE_DATE_PARAM', '01-01-2025').freeze
+
     class << self
       def office_details(office_code)
         query(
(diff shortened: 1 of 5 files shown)


robinsk1 (approved):

app/services/provider_data/provider_data_api_client.rb
```diff
@@ -19,10 +21,13 @@ def office_details(office_code)
...
 
       def contracted_office_details(office_code)
+        # effective_date only used in UAT environment
+        effective_date = HostEnv.uat? ? PROVIDER_API_EFFECTIVE_DATE_PARAM : nil
         # :nocov: Querying an external API
         params = {
-          'areaOfLaw' => 'CRIME LOWER'
-        }
+          'areaOfLaw' => 'CRIME LOWER',
+          'effectiveDate' => effective_date
```
robinsk1, line 29: should we send this param through if it is nil ie non-uat? maybe merge in the param if it is UAT instead?

ivanELEC: the compact method ensures that if that value is nil the key gets removed so effectively it never gets send through if it's non-uat (shown in tests)

PR #1252, opened 2026-06-04
CRM457-3068: Payment hub - Add 'Submission month' field to UI

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3068)

## Notes for reviewer

- [x] Add form components for: Original submission month (selectively show for NSM supplemental/appeal/amendment)
- [x] Include error handling/messaging
- [x] Push Original submission month to app store 
- [x] Show Original submission month details in CYA page and View payment page 
- [x] Handle defaulting to current date when not an unlinked NSM supplemental/appeal/amendment
- [x] Add tests

Extra additions:

- Add solicitor firm name to Claim details card
- Add consistency with prototype design for Solicitor Office Code and Firm Name on Claim details tab

### After changes:
image


app/controllers/payments/steps/nsm/claim_details_controller.rb
```diff
@@ -7,8 +7,35 @@ def edit
...
+        def form_key
+          'payments_steps_nsm_claim_detail_form'
+        end
+
+        def store_submission_date
+          return unless submission_date_present?
+
+          params[form_key]['original_submission_year'] = submission_year
+          params[form_key]['original_submission_month'] = submission_month
+          params[form_key]['original_submission_date'] = 1
```
robinsk1: 1? do you not just need to store year and month here?

app/controllers/payments/steps/nsm/claim_details_controller.rb
```diff
@@ -7,8 +7,35 @@ def edit
...
+        end
+
+        def submission_date_present?
```
robinsk1, line 27: should this and linked changes here not be handled by the form object - ie a validation on save?

app/forms/payments/steps/nsm/claim_detail_form.rb
```diff
@@ -50,16 +79,39 @@ def save
...
-          attributes.except('court_name_suggestion').each do |k, v|
-            multi_step_form_session[k.to_sym] = v
-          end
+          self.original_submission_year = Date.current.year
+          self.original_submission_month = Date.current.month
+        end
 
-          true
+        def valid_date?
+          Date.valid_date?(original_submission_year, original_submission_month, 1)
```
cmrgov, line 97: This still accepts year 101 because `Date.valid_date?(101, 4, 1)` is true. Can we validate a sensible year format/range before saving so the ticket's invalid-year case is blocked?

app/controllers/payments/steps/nsm/claim_details_controller.rb
```diff
@@ -7,8 +7,35 @@ def edit
         end
 
         def update
+          store_submission_date
```
robinsk1, line 10: might be form objects responsibility to do this stuff no?

app/view_models/payments/nsm_claim_details.rb
```diff
@@ -106,5 +108,9 @@ def submission_id
     def original_claim_row
       original_claim ? [table_heading('original_claim'), { text: original_claim, numeric: false }] : nil
     end
+
+    def original_submission_date
+      Date.new(@payable_claim['original_submission_year'].to_i, @payable_claim['original_submission_month'].to_i)
```
cmrgov, line 113: This will raise for old payable claims where month/year are nil, because nil becomes 0 and `Date.new(0, 0)` is invalid. Can we backfill/default these values or hide/fallback the row when they are absent?

app/controllers/payments/steps/nsm/claim_details_controller.rb
ivanELEC: This code isn't for validation, it's purely for the form object to be able to consume the params 
Because of the brackets in the param in this particular govuk_date_field component from the Gov UK Form Builder, when ActiveModel tries to create an attribute for it, it fails, because attributes can't have symbols in them 
So to allow the data to be parsed into the ActiveModel class, I need them renamed 

This is a problem for any instance of using govuk_date_field - tbh a better way of dealing with it would probably to have some form of application helper method that converts the keys so that any instance of this component can be dealt with but thought I'd do the basic implementation first and then improve

app/view_models/payments/nsm_claim_details.rb
ivanELEC: These two fields are now required fields for an Nsm payment (they will be and need to be populated in every Nsm scenario), I don't want to hide/fallback the row if the field isn't absent as when they should always exist for the user (if we did, it would be hiding a problem). 

Defaulting the data is risky also for the same reasons 

In app store, to allow people to use test claims created before this change I can backfill records created before the date I merge this in
I also will update the payment request seed file

ivanELEC: I realised that the "submission_date_needed?" method might be misleading here so renamed it to "new_submission_date_needed?"
This is basically just for indicating whether a new value for the submission date needs to be set (if it's not a "Create a new record" NSM appeal, amendment or supplemental - in those cases it defaults to the current year/month.

app/forms/payments/steps/nsm/claim_detail_form.rb
```diff
@@ -50,16 +79,39 @@ def save
...
-          return false unless valid?
+        def handle_original_submission_date
+          return if submission_date_needed?
 
-          attributes.except('court_name_suggestion').each do |k, v|
-            multi_step_form_session[k.to_sym] = v
-          end
+          self.original_submission_year = Date.current.year
+          self.original_submission_month = Date.current.month
+        end
```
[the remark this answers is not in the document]
ivanELEC: The review is inaccurate here - original_submission_year/month only gets reset if new_submission_date_needed? (renamed after commit) is true. Also we don't persist data to be saved and continued another day so a scenario where it's initially one date and changed due to revisiting the page isn't possible

app/presenters/payments/check_your_answers/claim_details_card.rb
```diff
@@ -135,6 +151,10 @@ def read_only?
 
       private
 
+      def original_submission_date
+        Date.new(session_answers['original_submission_year'].to_i, session_answers['original_submission_month'].to_i)
+      end
```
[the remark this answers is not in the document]
ivanELEC: We don't want to rescue the error here or have expectations for these fields to be empty because all NSM payments should have an original submission year

app/forms/payments/steps/nsm/claim_detail_form.rb
```diff
@@ -50,16 +80,42 @@ def save
...
+
+        def submission_date_must_be_present
+          return unless original_submission_year.blank? || original_submission_month.blank?
+
+          errors.add(:original_submission_date, :blank)
+        end
+
+        def submission_date_must_be_valid
+          errors.add(:original_submission_date, :invalid) unless valid_date?
+        end
```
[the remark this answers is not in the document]
ivanELEC: this is handled in the order that the validations are performed in

app/view_models/payments/nsm_claim_details.rb
```diff
@@ -106,5 +108,9 @@ def submission_id
...
+    end
```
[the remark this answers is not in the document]
ivanELEC: We shouldn't expect the original_submission_month and original_submission_year fields to ever be blank - if they are it means something went wrong and the page should indeed error out

cmrgov (dismissed):

app/controllers/payments/steps/nsm/claim_details_controller.rb
robinsk1: is govuk_date_field best way to implement this as we are storing the 'date' as an integer in app store - like picking/validating from array 1-12 for month or current year minus x number of years as an array, would avoid all the date to integer stuff ie capture it as an integer via the UI - its no longer a date, its an integer representing a month and an integer representing a year

ivanELEC: feels like a design question - would prefer that to be considered with UI/UX, to my knowledge there isn't a GDS component with this exact config (though could have a fieldset with a text input and a select box) 

Tbf I was actually in 2 minds about recording this data as 2 integers or just recording a full date. I mainly didn't just because the day is redundant in this case and I didnt want to have redundant data in app store.

cmrgov (approved):

app/controllers/payments/steps/nsm/claim_details_controller.rb
robinsk1: yeah, happy to go with what your proposing as you probably mulled over pros and cons already - i was just thinking its specifically not a `date` thats being recorded here its a year_month_period or something - this is actually quite tricky come to think of it. We don't want the redundant day but having two integer columns instead of one representing a singular attribute of information is also tricky. If its likely someone someday asks us they want to also store the day of submission then we'd need to add a day column with all the kerfuffle of backfilling etc that would cause - maybe just take the hit and store it as 1 date column with `1` as default day and live with the redundancy 😅 ?

PR #1264, opened 2026-06-15
CRM457-3068 Part 2: Use 'Month original claim assessed' as label for original_submission_date

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-2068)

## Notes for reviewer
In line with conversation in [this thread](https://mojdt.slack.com/archives/C034YPREUSF/p1781189511374749)

- Update label content for field 
- Only show Month original claim assessed field in CYA for non original NSM payments
- Update hint text in payments to use full stops
- Update error handling and error messages


app/controllers/payments/steps/nsm/claim_details_controller.rb
```diff
@@ -18,25 +18,11 @@ def form_key
         end
 
         def store_submission_date
-          return unless submission_date_present?
-
           # rubocop:disable Rails/StrongParametersExpect
-          params[form_key]['original_submission_year'] = submission_year
-          params[form_key]['original_submission_month'] = submission_month
+          params[form_key]['original_submission_year'] = params.dig(form_key, 'original_submission_date(1i)')
```
robinsk1, line 22: maybe another ticket but i'm leaning more to storing this in app-store as a date - might be easier to index/report on a date field

app/forms/payments/steps/nsm/claim_detail_form.rb
```diff
@@ -37,8 +37,9 @@ class ClaimDetailForm < BasePaymentsForm
         validates :ufn, presence: true, ufn: true
         validates :date_completed, :date_claim_assessed,
                   presence: true, multiparam_date: { allow_past: true, allow_future: false }
-        validate :submission_date_must_be_present, :submission_date_must_be_valid, :submission_date_must_be_in_past,
-                 if: :new_submission_date_needed?
+        validate  :submission_year_must_be_present, :submission_month_must_be_present, :submission_date_must_be_present,
```
robinsk1, line 40: more code/validation storing it as separate integers

ivanELEC: Yeah true sorting and indexing probably would be better off with a single date field as you mentioned in the other comment
Validation is a bit of a different story though (if we keep the same UI elements), casting as a date field within this form object would mean you can't independantly validate month and year, which is required as per the [error message specs Laurence gave us ](https://justiceuk.sharepoint.com/:x:/r/sites/NSCC/_layouts/15/Doc.aspx?sourcedoc=%7B0DB226BF-D5A9-4A87-9075-7B4ABBFC3FE9%7D&file=CRM-457%20error%20messages%2C%20provider%20claim%20a%20non-standard%20magistrates%20fee.xlsx&action=default&mobileredirect=true&DefaultItemOpen=1&wdOrigin=WAC.EXCEL.HOME-BUTTON%2CAPPHOME-WEB.FILEBROWSER.RECENT&wdPreviousSession=27fd3c72-a9ed-4cbb-93c4-76a0e4504913&wdPreviousSessionSrc=AppHomeWeb&ct=1781544723146) 

For example, if you put an invalid month into a date object, the whole object just fails to save and is invalid but there's no method to check if it's specifically the month that's invalid, so you either have a generic error message or you have to handle it as a string and then split the month out to check it anyways 

I would say we handle as 2 int fields in the form, but then store as a date when persisting to app store (and parse it as a date when presenting the data in the view payment page) to get the best of both worlds

robinsk1: yeah, that sounds better, ints and date to store - validations could be simply ranges (1-12 and current year minus 100 or something)

robinsk1 (approved):

PR #1276, opened 2026-06-24
Proof of Concept for Metabase API Export

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3128)

To test: 

1. Sign in as a supervisor 
2. Click on "Metabase test" 
3. Instructions are on the page


ivanELEC:
This likely does not work in the remote dev instance due to network policies in the crime application store namespace, I have feature flagged this code with the intent to temporarily merge it in, test in UAT and revert once confirmed whether or not it's the network policy that's blocking the request

app/controllers/metabases_controller.rb
```diff
@@ -0,0 +1,44 @@
+class MetabasesController < ApplicationController
```
robinsk1, line 1: might want to rename this as something more specific ie MetabaseDownload to prevent scope creep

app/controllers/metabases_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  before_action :authorize_supervisor
+
+  # :nocov:
+  def show
+    dashboard_ids = [47]
+    @iframe_urls = generate_metabase_urls(dashboard_ids)
+    return unless params[:card_id].present? && params[:start_date].present? && params[:end_date].present?
```
robinsk1, line 8: possibly want to have an allowlist for params

robinsk1 (approved):

robinsk1 (approved):

app/controllers/metabases_controller.rb
ivanELEC: will rectify after PoC

app/controllers/metabases_controller.rb
ivanELEC: will rectify after PoC

PR #1301, opened 2026-07-20
CRM457-3126:  Automate process of sending Emails containing CRM7 / CRM8 Finance reports

ivanELEC:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3126)

## Notes for reviewer

- [x] Add client for generating dashboard report downloads
- [x] Add mailler for generating email to finance team
- [x] Attach dashboard report to mailler
- [x] Schedule mailler
- [x] Add tests


app/jobs/payments/schedule_report_base.rb
```diff
@@ -0,0 +1,40 @@
...
+      Rails.logger.info 'Cleared tmp/uploaded/reports directory'
+    end
+    # rubocop:enable Metrics/AbcSize
+
+    def recipient_key
+      raise NotImplementedError, "#{self.class} has not implemented method '#{__method__}'"
+    end
+
+    def recipients
+      ENV.fetch(recipient_key, '').split(',').map(&:strip).compact_blank
```
robinsk1, line 25: why do we have recipients as ENV variables?

robinsk1 (commented):
LGTM ...but i have big reservations about how we are fetching this data here ie `assess -> metabase -> app-store -> metabase -> assess` when it should just be sending an email from  `appstore`. It seems brittle, metabase doesn't really need to be here at all. I could be missing something here though, happy to wind back on that

app/jobs/payments/schedule_report_base.rb
robinsk1: could have had something like 'receives emai' in the CW admin panel

robinsk1 (approved):

app/jobs/payments/schedule_report_base.rb
ivanELEC: they need to be secrets due to GDPR as some are individuals work emails, so they will either need to be secrets or in DB, could add to admin panel and move to DB as a post MVP thing but also I wonder if that page will even persist once Assess moves to SILAS

ivanELEC:
> LGTM ...but i have big reservations about how we are fetching this data here ie `assess -> metabase -> app-store -> metabase -> assess` when it should just be sending an email from `appstore`. It seems brittle, metabase doesn't really need to be here at all. I could be missing something here though, happy to wind back on that

yeah I see your point, I think we should discuss this again. The main reason for Metabase here is to enable us to have agility on changes on data transform in case some more complicated summarisations arise e.g. if we need to use CountIf aggregates etc, that said: 
1. The requirements seem quite fixed, there are no complex additions to the reporting on the horizon
2. We can still utilise Metabase for that by generating what we need in a question and converting that to SQL, then putting it into our DB as a view (that's our current process anyways minus the creating the view part)

Taking Metabase out of the equation for production data would decrease our threat scores etc. Maybe we keep this in until a successful calendar month finance process is complete, work out any kinks in the data and then switch to a more baked in API?

PR #1306, opened 2026-07-23
CRM457-3215 modify policy and add more tests

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3215)

### DIFF
--- a/app/policies/payment_policy.rb
+++ b/app/policies/payment_policy.rb
@@ -1,23 +1,19 @@
 class PaymentPolicy < ApplicationPolicy
   def show?
-    service_access? && user_access?
+    permitted?
   end
 
   def index?
-    service_access? && user_access?
+    permitted?
   end
 
   def update?
-    service_access? && user_access?
+    permitted?
   end
 
   private
 
-  def service_access?
-    user.nsm_access? || user.pa_access?
-  end
-
-  def user_access?
-    user.caseworker? || user.supervisor?
+  def permitted?
+    user.supervisor? || (user.caseworker? && user.nsm_access?)
   end
 end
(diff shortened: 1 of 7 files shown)


app/policies/payment_policy.rb
```diff
@@ -1,23 +1,19 @@
...
   private
 
-  def service_access?
-    user.nsm_access? || user.pa_access?
-  end
-
-  def user_access?
-    user.caseworker? || user.supervisor?
+  def permitted?
+    user.supervisor? || (user.caseworker? && user.nsm_access?)
```
cmrgov: These checks look across all of the user’s roles. Does that mean caseworker/pa plus viewer/nsm would pass, even though the Caseworker role does not have NSM or All access?

Can we check the role and service on the same record and add a mixed-role policy spec?

ivanELEC: Hmm that's a good question - Really and truly the way we've written roles it should only be one role per person but the DB structure and how roles generally work allow for multiple. I think this depends on whether roles are additive or not. I think if roles are additive, the way this is already written is fine

robinsk1: Good catch —  the previous check could mix permissions across different role records ( caseworker/pa  +  viewer/nsm ) and allows access.

Updated PaymentPolicy so  access is now only for supervisors, or caseworkers with  nsm / all  on that specific role assignment

cmrgov: Yep the payment routes are now protected, but the NSM claim-details and decision pages might still show “Create payment request” based only on claim eligibility?

A viewer with NSM access could still view those pages, so do we need to gate both links with `policy(:payment).show?` and add corresponding viewer/PA-only spec?

cmrgov (approved):

PR #1318, opened 2026-08-03
CRM457-3245 fix work-items xml uploads

robinsk1:
## Description of change

[Link to relevant ticket](https://dsdmoj.atlassian.net/browse/CRM457-3245)

### DIFF
--- a/app/view_models/base_view_model.rb
+++ b/app/view_models/base_view_model.rb
@@ -10,7 +10,8 @@ class BaseViewModel
       @klass = "#{submission.namespace}::V#{submission.json_schema_version}::#{class_type.to_s.camelcase}".constantize
       @submission = submission
       if nesting.any?
-        @rows = submission.data.dig(*nesting)
+        nested_rows = submission.data.dig(*nesting)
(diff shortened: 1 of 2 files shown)


app/view_models/base_view_model.rb
```diff
@@ -10,7 +10,8 @@ def initialize(class_type, submission, *nesting)
       @klass = "#{submission.namespace}::V#{submission.json_schema_version}::#{class_type.to_s.camelcase}".constantize
       @submission = submission
       if nesting.any?
-        @rows = submission.data.dig(*nesting)
+        nested_rows = submission.data.dig(*nesting)
+        @rows = nested_rows.is_a?(Array) || nested_rows.nil? ? nested_rows : [nested_rows]
```
ivanELEC, line 14: Would be preferable not to change the contract of the base view model of a claim. 
We wouldn't expect our service to generate a claim hash that has data structured like this (even if there's only one element there would be a single element array) - so would prefer if this could be solved at the point where Nokigiri is processing the XML file if possible

PR #1323, opened 2026-08-05
CRM457-3186: Add Assess switchable auth strategy groundwork for SiLAS

cmrgov:
## Description of change

[CRM457-3186: Add Assess switchable auth strategy groundwork for SiLAS](https://dsdmoj.atlassian.net/browse/CRM457-3186)

Adds the Assess-side groundwork to switch between the current Azure (`azure_ad`) auth path and a future SiLAS path, without committing to SiLAS yet. `ASSESS_AUTH_PROVIDER` selects the provider and defaults to `azure_ad`.

- Keeps Azure as the default; `dev_auth` still maps to the Azure strategy for local sign-in.
- Routes both providers through a common `Auth::UserAuthenticator` and strategy interface.
- Stores SiLAS identity and roles separately (`authentication_identities`, `silas_roles`) so Azure rollback stays possible.
- Reads the SiLAS claim as `USER_NAME`, `USER_EMAIL`, and `LAA_APP_ROLES`, and supports SiLAS-shaped claims through DevAuth for local testing.
- Fails closed for missing, malformed, or unknown claims and roles, and for unknown or deactivated users (no auto-provisioning).
- Maps `LAA_APP_ROLES` values to Assess `{role_type, service}` roles via `SILAS_ROLE_MAPPINGS`, and uses those roles in the existing Pundit policies when SiLAS is selected.
- Makes user and role management read-only while SiLAS is the provider.

## Notes for reviewer

Still a draft, not ready to enable or merge. The deployed provider stays `azure_ad` and deployed `SILAS_ROLE_MAPPINGS` stays empty.

Had the initial Assess <> SiLAS contract call with Callum Brett (SiLAS) on 06 Aug. Most of the previously open questions are settled, and the code already matches what was agreed.

Confirmed:
- **Stable identifier is `USER_NAME`** (the SiLAS `legacyUserID`). The strategy links on it via `authentication_identities.subject`.
- **Claims are `USER_NAME`, `USER_EMAIL`, `LAA_APP_ROLES`** (optional `LAA_ACCOUNTS` for office codes). `first_name`/`last_name` come from the base Entra claim, not SiLAS enrichment — the strategy already falls back accordingly.
- **Roles flatten to combined `role_type + service` values** (e.g. caseworker + NSM), which is what `SILAS_ROLE_MAPPINGS` expects. A later "Request for Payment" role can be added without reworking this.
- **Backfill is by email:** existing users get a `silas` identity row keyed on `USER_NAME`, matched from a users/roles export (Callum can generate the SiLAS SQL import). Tracked under CRM457-3187.
- **Environments:** Dev stays on the bypass (DevAuth); first real SiLAS login is NLE attached to our UAT, then prod. Entra app registration and claim enrichment are owned by EUCS IdAM (Terraform for NLE/prod); SiLAS service and roles are loaded via the SiLAS admin team.

Still open before this leaves draft:
- Callum's final `LAA_APP_ROLES` value strings, to populate `SILAS_ROLE_MAPPINGS`.
- Supervisor access scope for dashboards and payments.
- Whether we need an email-domain/app/user allow-list on top of fail-closed.

Contract ticket CRM457-3191; rollout/backfill ticket CRM457-3187.

Pending: the read-only SiLAS user-management state needs a screenshot before this leaves draft.

Set `ASSESS_AUTH_PROVIDER=silas` locally, pre-provision a user with a `silas` identity and a `USER_NAME`, then sign in through DevAuth with SiLAS-shaped claims. Confirm the mapped SiLAS roles sync and that user management is read-only.

Switch back to `azure_ad` and confirm the existing Azure sign-in and local role management still work.


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -3,13 +3,11 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+      authenticate_from_omniauth
+    end
 
-      if @user
-        sign_in_and_redirect @user, event: :authentication
-      else
-        throw(:warden, recall: 'Errors#forbidden', message: :forbidden)
-      end
+    def silas
+      authenticate_from_omniauth
```
robinsk1: these call the same method - why the split if its either `silas` or `azure` thats called up the chain its the same call to the same private method?

app/lib/omni_auth/strategies/dev_auth.rb
```diff
@@ -50,6 +54,34 @@ def last_name
...
+        }
+      end
+
+      def simulated_silas_roles
+        return [] unless user
+
+        user.silas_roles.presence || user.roles
+      end
+
+      def simulated_silas_claim_values
```
robinsk1: not sure what simulated means here/why simulated

app/models/authentication_identity.rb
```diff
@@ -0,0 +1,14 @@
+class AuthenticationIdentity < ApplicationRecord
+  PROVIDERS = %w[azure_ad silas].freeze
+
+  belongs_to :user
+
+  validates :provider, inclusion: { in: PROVIDERS }, uniqueness: { scope: :user_id }
+  validates :subject, presence: true, uniqueness: { scope: :provider }
+
+  def authentication_expired?
+    return true unless last_authenticated_at
```
robinsk1: does that work for new users also?

app/policies/claim_policy.rb
```diff
@@ -1,18 +1,18 @@
 class ClaimPolicy < ApplicationPolicy
   def update?
-    service_access? && !record.closed? && record.assigned_user_id == user.id
+    service_access? && !record.closed? && record.assigned_user_id == user.id && !user.viewer?(:nsm)
```
robinsk1: why changes in this policy?

app/services/auth/strategies/azure_ad.rb
```diff
@@ -0,0 +1,64 @@
...
+      def initialize(auth_hash)
+        @auth_hash = auth_hash
+      end
+
+      def call
+        user = local_user
+        return failure(:not_authorized) unless user
+
+        sync_user!(user)
+        Result.new(user: user, failure_reason: nil)
```
robinsk1: `Result` is quite generic naming wise? is that a new class?

app/services/authorization/role_sources/database.rb
```diff
@@ -0,0 +1,21 @@
+module Authorization
+  module RoleSources
+    class Database
```
robinsk1: what does `Database` mean here?

app/lib/auth/provider.rb
```diff
@@ -0,0 +1,60 @@
...
+    CALLBACK_ALIASES = { 'dev_auth' => 'azure_ad' }.freeze
+
+    class UnknownProvider < StandardError; end
+
+    class << self
+      def current
+        fetch(ENV.fetch('ASSESS_AUTH_PROVIDER', 'azure_ad'))
+      end
+
+      def fetch(name)
```
ivanELEC: nitpick: maybe we use a different name than fetch to not mix it up with the existing method for hashes. Feels more like `construct` or `build`. Not incredibly important though

spec/requests/payments/courts_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe 'Payment courts' do
```
robinsk1: odd how this has crept in

spec/services/auth/silas_role_mapper_spec.rb
```diff
@@ -0,0 +1,164 @@
+require 'rails_helper'
+
+RSpec.describe Auth::SilasRoleMapper do
+  let(:configured_mappings) do
+    {
+      'Assess Caseworker (All)' => { role_type: 'caseworker', service: 'all' },
```
robinsk1: `Assess Caseworker (All)` etc seems like an odd string format/pattern for a key - is there a reason for that format?

robinsk1 (dismissed):

app/lib/auth/provider.rb
cmrgov: Renamed to build

app/controllers/users/omniauth_callbacks_controller.rb
cmrgov: Devise routes callbacks to actions named after each provider, so both entry points are needed. They share the authentication method, which selects the strategy and checks the configured provider. Added a short comment to explain the separate actions.

app/lib/omni_auth/strategies/dev_auth.rb
cmrgov: This is the local DevAuth bypass. It builds SiLAS-shaped claims from stored roles so we can test locally without calling SiLAS. It uses the SiLAS snapshot when present, otherwise local roles. Added a comment to make that clear.

app/models/authentication_identity.rb
cmrgov: Yes for first authentication of an existing Assess account. A missing timestamp means authentication is required; the strategy sets it when sign-in succeeds. SiLAS links the existing account by email on first sign-in. It does not create an Assess account for an unknown email. Both missing timestamps and unknown users are covered by specs.

app/policies/claim_policy.rb
cmrgov: The viewer check now applies to the service being assessed. Someone can be an NSM caseworker and a PA viewer, so a global viewer check would block legitimate NSM work. The update check also prevents an existing assignment from letting a viewer edit that service’s records. Specs cover both mixed-role directions.

app/services/auth/strategies/azure_ad.rb
cmrgov: Yes, this is the new Auth::Result value object. It carries the user or failure reason and exposes success? for both providers. I've made Auth::Result explicit at each call site so its purpose is clearer.

app/services/authorization/role_sources/database.rb
cmrgov: Renamed this to StoredRoles. It reads roles stored in Assess: user.roles for Azure and user.silas_roles for the SiLAS snapshot. The SiLAS subclass checks snapshot freshness. It doesn't query SiLAS during permission checks.

spec/services/auth/silas_role_mapper_spec.rb
cmrgov: Yes - this was a constraint from the SILAS side. Their keys are also the human readable strings in their GUI. To split out parameterised keys from the human readable strings was going to be a big ask on their side so we opted with accepting this format 🤷

spec/requests/payments/courts_spec.rb
cmrgov: removed

robinsk1 (dismissed):

helm_deploy/templates/auth-configuration.yaml
```diff
@@ -0,0 +1,11 @@
...
+{{- if not (lookup "v1" "ConfigMap" .Release.Namespace $name) -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ $name }}
+  annotations:
+    helm.sh/resource-policy: keep
+data:
+  auth_provider: azure_ad
+{{ end }}
```
ivanELEC: the "configuration" configmap in this folder is already available for settings on the app, I don't think we need to create a separate configmap, we could just add an extra data key to the existing one

robinsk1 (approved):

ivanELEC (approved):

